PR #32, opened 2026-01-26
feat: Add multi-package install and uninstall support

### LINKED ISSUE #18: Install multiple doesn't seem to work
When trying to do multi-install, the second argument is seen as a separate argument

```
zb install biome ffmpeg
error: unexpected argument 'ffmpeg' found

Usage: zb install [OPTIONS] <FORMULA>

For more information, try '--help'.
```

Probably the Install command needs `formula` to be a `Vec<String>` ...and all of the resolution logic needs to support multiple roots... - one could do a quick hack and just run them in serial, but I'm not sure that doing it properly is so much more work that the quick hack makes sense

vyrti:
## Summary
- Adds support for installing and uninstalling multiple packages in a single command
- Users can now run `zb install biome ffmpeg` instead of running separate commands

Fixes #18

## Test plan
- [ ] Test `zb install pkg1 pkg2` installs both packages
- [ ] Test `zb uninstall pkg1 pkg2` uninstalls both packages
- [ ] Test single package install/uninstall still works

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,12 @@ This leads to dramatic speedups, up to 5x cold and 20x warm. Full benchmarks [he
 ##  Using `zb`
 
 ```bash
-zb install jq        # install jq
-zb install wget git  # install multiple
-zb uninstall jq      # uninstall
-zb reset         # uninstall everything
-zb gc                # garbage collect unused store entries
+zb install jq           # install jq
+zb install wget git     # install multiple
+zb uninstall jq         # uninstall jq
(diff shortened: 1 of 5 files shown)


vyrti:
I understood that this https://github.com/lucasgelfond/zerobrew/pull/21 only cover installation but it does not make sense, as we need to uninstall several at once as well

this also pass clippy, fmt

lucasgelfond:
pls rebase

vyrti:
@lucasgelfond done

PR #41, opened 2026-01-26
feat: tradeoff / complex: Improve fuzzy matching to include prefix/substring matches and FULL cache implementation

vyrti:
## Summary

- Improve fuzzy matching to include prefix/substring matches (e.g., `aws-sso` matches `aws-sso-cli`)
- **Cache formula names list** for efficient fuzzy matching with 24-hour TTL
- Add `zb update` command to force-refresh the formula list cache

### Fuzzy Matching Improvements
- Priority 0: Exact prefix match (e.g., `aws-sso` matches `aws-sso-cli`)
- Priority 1: Query is substring (e.g., `sso` matches `aws-sso-cli`)
- Priority 2: Close Levenshtein distance for typo correction

### Formula Names Caching
- Add `get_if_fresh()` method to `ApiCache` for TTL-based cache lookups
- Add caching to `get_formula_names()` with 24-hour TTL (like Homebrew's default)
- Use conditional HTTP requests (ETag/Last-Modified) to minimize bandwidth
- Wire up cache in CLI for suggestions on missing formula errors

| Before | After |
|--------|-------|
| ~29.6 MB network per typo | 0 bytes (cache hit) |
| No offline support | Works offline with cached data |
| No update command | `zb update` force-refreshes cache |

## Test plan

- [x] Run `zb install nonexistent-pkg` - should download formula list, cache it, show suggestions
- [x] Run `zb install another-typo` - should be instant (no network for list)
- [x] Run `zb update` - should force-refresh the formula list cache
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/zb_io/src/api.rs
+++ b/zb_io/src/api.rs
@@ -125,7 +125,10 @@ impl ApiClient {
     ///
     /// - If `force_refresh` is false and a fresh cache exists (within 24h), returns cached data.
     /// - Otherwise, performs a conditional HTTP request (with ETag/Last-Modified) and updates cache.
-    pub async fn get_formula_names_cached(&self, force_refresh: bool) -> Result<Vec<String>, Error> {
+    pub async fn get_formula_names_cached(
+        &self,
+        force_refresh: bool,
+    ) -> Result<Vec<String>, Error> {
         let url = format!("{}.json", self.base_url);
 
         // Check for fresh cache (unless force refresh)
--- a/zb_cli/src/main.rs
+++ b/zb_cli/src/main.rs
@@ -350,10 +350,7 @@ async fn run(cli: Cli) -> Result<(), zb_core::Error> {
 
     // Handle Update command early - it only needs cache, not full installer
     if matches!(cli.command, Commands::Update) {
-        println!(
-            "{} Updating formula list...",
-            style("==>").cyan().bold()
-        );
+        println!("{} Updating formula list...", style("==>").cyan().bold());
 
         // Ensure cache directory exists
         let cache_dir = cli.root.join("cache");


zb_io/src/api.rs
lucasgelfond, line 1: this seems fairly wasteful - we fetch the entire list of brew packages each time? is this how brew itself does it / is there some more efficient way for us to do this?

vyrti: brew have 
**brew update** 
and keeping cache locally for 24h as its 24mb each time

so I will update to do the same
but first "install" request will still download 24mb, so either skip this fuzzy / search kind of things completely or loose time on first request. This is tradeoff for you to decide :)

vyrti: @lucasgelfond updated. But basically this means introduction of full cache system like in brew, that downloads cache when fuzzy logic triggered for the first time and will keep it for 24h, so all other requests will go to cache as well.. so this is not just fuzzy, its about do we want caching like in brew or not? i believe things like fuzzy and search will actually benefit from it, but we will loose speed on first request (only on first one)
and if we want to implement a cache then this PR needs more code / tests / attention before merging
like cache does not being triggered now on usual situation, how it will work / update after 24h etc etc etc

lucasgelfond: oh wait - I didn't realize the whole package dir was just 24mb once - i'm okay to keep forever? 

lose speed on first request, sounds good 

my sense would be though, if we are downloading the package dir, we'll still need to deal with cache invalidation / all of this annoying stuff? 

lets discuss, I can see merits to a lot of the approaches, in any sense i'm onboard with some version of this appraoch

vyrti: i think it might be worth considering https://github.com/lucasgelfond/zerobrew/pull/47 first
and then just update this PR to use that cache, for fuzzy, and fast local search will be possible then also as additional pr

Issue #40, opened 2026-01-26
Emacs doesn't work on macOS Sequoia


livcher:
While trying to reproduce #27, I faced a different issue:
```
❯ emacs                                                                                                                                                                            
dyld[66864]: Symbol not found: _posix_spawn_file_actions_addchdir
  Referenced from: <96A30EF7-FDD2-37F0-AA7B-B666089CD27D> /opt/zerobrew/prefix/Cellar/emacs/30.2/bin/emacs-30.2 (built for macOS 26.0 which is newer than running OS)
  Expected in:     <068263F7-CF15-3E1C-A305-5DF8F9C51AD5> /usr/lib/libSystem.B.dylib
[2]    66864 abort      emacs
```
Looks like zb is pulling the latest binary instead of the one that works on my system. I expect this would also be an issue for people on Intel macs (even with Tahoe) as they obviously have different builds.

This doesn't happen if I install emacs via Homebrew.

macOS Sequoia 15.7.3, M2, zsh

cachebag:
Related to #27

Issue #31, opened 2026-01-26
Homebrew placeholders not substituted


carterwsmith:
Some formulae install scripts that contain Homebrew placeholders like `@@HOMEBREW_CELLAR@@`. When installed with zerobrew these placeholders are not substituted during install.

This causes installed binaries to fail at execution time even though the package installs successfully.

Reproduce:
```
zb install mole
mo
```

Error:
```
/opt/zerobrew/prefix/bin/mo: line 10: @@HOMEBREW_CELLAR@@/mole/1.23.2/libexec/lib/core/common.sh: No such file or directory
```

PR #45, opened 2026-01-26
bugfix: tradeoff / complex: Patch hardcoded paths in binaries (BREAKING CHANGE)

### LINKED ISSUE #27: emacs: fails to start due to hard-coded /opt/homebrew paths
Hi,

The emacs homebrew package seems to have hard-coded `/opt/homebrew` somewhere, so it fails to start at runtime with an error:
```
➜  ~ which emacs
/opt/zerobrew/prefix/bin/emacs
➜  ~ emacs
Warning: arch-dependent data dir '/opt/homebrew/Cellar/emacs/30.2_2/libexec/emacs/30.2/aarch64-apple-darwin25.1.0/': No such file or directory
Warning: arch-independent data dir '/opt/homebrew/Cellar/emacs/30.2_2/share/emacs/30.2/etc/': No such file or directory
Warning: Lisp directory '/opt/homebrew/Cellar/emacs/30.2_2/share/emacs/30.2/lisp': No such file or directory
Error: /opt/homebrew/Cellar/emacs/30.2_2/share/emacs/30.2/etc/charsets: No such file or directory
Emacs will not function correctly without the character map files.
Please check your installation!
```

vyrti:
## ⚠️ BREAKING CHANGE ⚠️

**This PR changes zerobrew's installation path structure. Existing installations must be reset.**

### Migration Required
```bash
zb reset -y
# Then reinstall packages
```

### The Problem

Emacs (and other packages) have **hardcoded paths embedded as string constants** in their compiled binaries. When running emacs installed by zerobrew:

```
Warning: arch-dependent data dir '/opt/homebrew/Cellar/emacs/30.2_2/libexec/emacs/30.2/aarch64-apple-darwin24.6.0/': No such file or directory
Warning: arch-independent data dir '/opt/homebrew/Cellar/emacs/30.2_2/share/emacs/30.2/etc/': No such file or directory
Warning: Lisp directory '/opt/homebrew/Cellar/emacs/30.2_2/share/emacs/30.2/lisp': No such file or directory
```

The binary was compiled with `/opt/homebrew/Cellar/...` hardcoded, but zerobrew was installing to `/opt/zerobrew/prefix/Cellar/...`.

### Why Binary Patching Wasn't Working

Binary string replacement requires the new string to be **the same length or shorter** (padded with null bytes):

| Path | Length |
|------|--------|
| `/opt/homebrew` | 13 chars |
| `/opt/zerobrew/prefix` | 20 chars |
| `/opt/homebrew/Cellar` | 20 chars |
| `/opt/zerobrew/prefix/Cellar` | 27 chars |

**The zerobrew paths were 7 characters longer than Homebrew paths!** This made binary patching impossible.

### The Solution

Remove the `/prefix/` level from zerobrew's directory structure:

| Before | After |
|--------|-------|
| `/opt/zerobrew/prefix/Cellar/...` | `/opt/zerobrew/Cellar/...` |
| `/opt/zerobrew/prefix/bin/...` | `/opt/zerobrew/bin/...` |

Now the paths match exactly:
- `/opt/zerobrew` = 13 chars (matches `/opt/homebrew`)
- `/opt/zerobrew/Cellar` = 20 chars (matches `/opt/homebrew/Cellar`)

### Changes

1. **`zb_cli/src/main.rs`**: Changed default `--prefix` from `/opt/zerobrew/prefix` to `/opt/zerobrew`

2. **`zb_io/src/materialize.rs`**: Added `patch_embedded_strings()` function that:
   - Searches Mach-O binaries for hardcoded `/opt/homebrew/Cellar/{pkg}/` paths
   - Replaces with the actual keg path (handles version mismatches like `30.2_2` → `30.2`)
   - Replaces remaining `/opt/homebrew` prefixes with `/opt/zerobrew`
   - Re-signs modified binaries with ad-hoc signatures

3. **`zb_core/src/bottle.rs`**: Added macOS version detection for bottle selection

4. **`install.sh`**: Updated all path references

### Additional Fix: Bottle/API Version Mismatch

A second issue was discovered where the **bottle tarball version differs from the API version**:
- Bottle contains version `30.2_2`
- API returns `rebuild=0`, causing zerobrew to use `30.2`

When patching paths like `/opt/homebrew/Cellar/emacs/30.2_2/share/...` to `/opt/zerobrew/Cellar/emacs/30.2/share/...`, the 2-character difference caused null-padding that **truncated the path**:

```
emacs/30.2\0\0/share/info/emacs  (null bytes break the path)
```

**Fix**: The installer now detects the actual version directory inside the extracted bottle and uses that version for the keg path, ensuring paths match exactly.

Changes:
- **`zb_io/src/extract.rs`**: Added `detect_bottle_version()` function
- **`zb_io/src/install.rs`**: Use detected bottle version instead of API's `effective_version()`
- **`zb_io/src/materialize.rs`**: Improved pdmp patching with buffer-based approach for robustness

### Testing

```bash
# After this PR
zb reset -y
zb install emacs
/opt/zerobrew/bin/emacs --version
# GNU Emacs 30.2 ✓

# Verify paths are correct (no null truncation)
strings /opt/zerobrew/Cellar/emacs/*/libexec/emacs/*/*.pdmp | grep "zerobrew/Cellar/emacs" | head -3
# Should show full paths like: /opt/zerobrew/Cellar/emacs/30.2_2/share/emacs/30.2/etc
```

Fixes #27 #31 #40

### DIFF
--- a/zb_io/src/install.rs
+++ b/zb_io/src/install.rs
@@ -300,17 +300,17 @@ impl Installer {
                         .unwrap_or_else(|| formula.effective_version());
 
                     // Materialize to cellar
-                    let keg_path = match self.cellar.materialize(
-                        &formula.name,
-                        &version,
-                        &store_entry,
-                    ) {
-                        Ok(path) => path,
-                        Err(e) => {
-                            error = Some(e);
-                            continue;
-                        }
-                    };
+                    let keg_path =
+                        match self
+                            .cellar
+                            .materialize(&formula.name, &version, &store_entry)
+                        {
+                            Ok(path) => path,
+                            Err(e) => {
+                                error = Some(e);
+                                continue;
+                            }
+                        };
 
                     report(InstallProgress::UnpackCompleted {
                         name: formula.name.clone(),
--- a/zb_io/src/materialize.rs
+++ b/zb_io/src/materialize.rs
@@ -539,10 +539,11 @@ fn patch_embedded_strings(
                 // Found the prefix, now find the end of the version (next / or null)
                 let version_start = i + old_prefix_bytes.len();
                 let mut version_end = version_start;
+                // Find end of version: stop at /, null, or space (configure flags)
                 while version_end < data.len()
                     && data[version_end] != b'/'
                     && data[version_end] != 0
-                    && data[version_end] != b' ' // Stop at space (used in configure flags)
+                    && data[version_end] != b' '
                 {
                     version_end += 1;
                 }
@@ -634,7 +635,7 @@ fn patch_embedded_strings(
         .into_iter()
         .filter_map(|e| e.ok())
         .filter(|e| e.file_type().is_file())
-        .filter(|e| e.path().extension().map_or(false, |ext| ext == "pdmp"))
+        .filter(|e| e.path().extension().is_some_and(|ext| ext == "pdmp"))
         .map(|e| e.path().to_path_buf())
         .collect();
 


vyrti:
It seems that another tradeoff is needed
Currently app uses /prefix

But to make bin releases from homebrew work we have 3 options:

The zerobrew paths are 7 characters longer than Homebrew paths, making binary patching impossible.                                                                                   
                                                                                                                                  
❯ 1. Option A: Remove /prefix/ (Recommended)                                                                                      
     Change /opt/zerobrew/prefix/Cellar to /opt/zerobrew/Cellar. Enables exact binary patching. Requires changing CLI defaults.   
  2. Option B: Homebrew symlinks                                                                                                  
     Create symlinks from /opt/homebrew/Cellar to zerobrew. No binary patching needed but may conflict with real Homebrew.        
  3. Option C: Shorter base path (/opt/zb)                                                                                        
     Use /opt/zb instead of /opt/zerobrew. Enables binary patching but uses non-standard path.
     
This current implementation uses 1, adds a macos version check as well, and emacs DOES show version at least now
Continue research
but i am unsure why we have prefix while homebrew does not, i mean a word in path, not meaning of it

vyrti:
emacs works now
and there are high chance this will help other apps to work as well

PR #49, opened 2026-01-27
feat: support http range requests to download different parts of the same file

### LINKED ISSUE #42: May want to consider parallelizing further with http range requests?
Many downloaded artifacts are quite large. Your parallelism will be reduced when there are only a few larger items remaining to download.

To maintain high parallelism through the full download it might be possible to use http range requests to download different parts of the same file in different http requests (I believe ghcr.io supports this).

Happy to PR this at some point if I find the time.

fatelei:
fix #42

### DIFF
--- a/zb_io/src/download.rs
+++ b/zb_io/src/download.rs
@@ -1,14 +1,16 @@
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::io::Write;
 use std::path::PathBuf;
 use std::sync::Arc;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::time::{Duration, Instant};
 
 use futures_util::StreamExt;
 use futures_util::future::select_all;
 use reqwest::StatusCode;
-use reqwest::header::{AUTHORIZATION, CONTENT_LENGTH, HeaderValue, WWW_AUTHENTICATE};
+use reqwest::header::{
+    ACCEPT_RANGES, AUTHORIZATION, CONTENT_LENGTH, CONTENT_RANGE, HeaderValue, WWW_AUTHENTICATE,
+};
 use serde::Deserialize;
 use sha2::{Digest, Sha256};
 use tokio::sync::{Mutex, Notify, RwLock, Semaphore, mpsc};
@@ -23,6 +25,57 @@ const RACING_CONNECTIONS: usize = 1;
 
 /// Delay between starting each racing connection (ms)
 const RACING_STAGGER_MS: u64 = 200;
+
+/// Minimum file size to use chunked downloads (10MB)
+const CHUNKED_DOWNLOAD_THRESHOLD: u64 = 10 * 1024 * 1024;
+
+/// Global download concurrency limit
+/// Total number of concurrent connections across all downloads to avoid
+/// overwhelming servers and the local network. Based on industry best practices
+/// (npm uses 20-50, we use a conservative 20 for HTTP/1.1 compatibility).
+const GLOBAL_DOWNLOAD_CONCURRENCY: usize = 20;
+
+/// Maximum concurrent chunk downloads per file
+/// Chosen to divide GLOBAL_DOWNLOAD_CONCURRENCY among multiple large file downloads.
+/// With 20 global concurrency, we can have 3-4 large files downloading concurrently.
+const MAX_CONCURRENT_CHUNKS: usize = 6;
+
+/// Maximum retry attempts for failed chunk downloads
+const MAX_CHUNK_RETRIES: u32 = 3;
+
+fn calculate_chunk_size(file_size: u64) -> u64 {
+    const MIN_CHUNK_SIZE: u64 = 5 * 1024 * 1024;
+    const MAX_CHUNK_SIZE: u64 = 20 * 1024 * 1024;
+
+    let target_chunks = MAX_CONCURRENT_CHUNKS as u64;
+    let chunk_size = file_size / target_chunks;
+
+    chunk_size.clamp(MIN_CHUNK_SIZE, MAX_CHUNK_SIZE)
+}
+
+/// Context for chunk download operations
+struct ChunkDownloadContext<'a> {
+    client: &'a reqwest::Client,
+    token_cache: &'a TokenCache,
+    url: &'a str,
+    progress: Option<DownloadProgressCallback>,
+    name: Option<String>,
+    file_size: u64,
+    total_downloaded: Arc<AtomicU64>,
+}
+
+/// Context for chunked download operations
+struct ChunkedDownloadContext<'a> {
+    blob_cache: &'a BlobCache,
+    client: &'a reqwest::Client,
+    token_cache: &'a TokenCache,
+    url: &'a str,
+    expected_sha256: &'a str,
+    name: Option<String>,
+    progress: Option<DownloadProgressCallback>,
+    file_size: u64,
+    global_semaphore: &'a Arc<Semaphore>,
+}
 
 /// Callback for download progress updates
 pub type DownloadProgressCallback = Arc<dyn Fn(InstallProgress) + Send + Sync>;
@@ -82,10 +135,15 @@ pub struct Downloader {
     client: reqwest::Client,
     blob_cache: BlobCache,
     token_cache: TokenCache,
+    global_semaphore: Option<Arc<Semaphore>>,
 }
 
 impl Downloader {
     pub fn new(blob_cache: BlobCache) -> Self {
+        Self::with_semaphore(blob_cache, None)
+    }
+
+    pub fn with_semaphore(blob_cache: BlobCache, semaphore: Option<Arc<Semaphore>>) -> Self {
         // Use HTTP/2 with connection pooling for better performance
         Self {
             client: reqwest::Client::builder()
@@ -102,6 +160,7 @@ impl Downloader {
                 .unwrap_or_else(|_| reqwest::Client::new()),
             blob_cache,
             token_cache: Arc::new(RwLock::new(HashMap::new())),
+            global_semaphore: semaphore,
         }
     }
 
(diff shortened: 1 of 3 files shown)


zb_io/src/download.rs
```diff
@@ -442,6 +495,178 @@ async fn fetch_bearer_token_internal(
...
+        cb(InstallProgress::DownloadStarted {
+            name: n.clone(),
+            total_bytes: Some(file_size),
+        });
+    }
+
+    let total_downloaded = Arc::new(AtomicU64::new(0));
+
+    let mut handles = Vec::new();
+    for chunk in chunks {
```
cachebag, line 762: For a 100MB file, this creates 20 concurrent connections (100MB / 5MB chunks). 

I think this PR is nice to have but maybe limit ourselves to 6 connections per HTTP/1.1 server for now? [Ref](https://textslashplain.com/2024/11/22/parallel-downloading/)

We could add a semaphore or use `tokio::task::JoinSet` with a max concurrency limit if you see this as reasonable.

zb_io/src/download.rs
```diff
@@ -442,6 +495,178 @@ async fn fetch_bearer_token_internal(
...
+                });
+            }
+
+            Ok::<(Vec<u8>, u64), Error>((chunk_data, chunk.offset))
+        });
+
+        handles.push(handle);
+    }
+
+    let mut chunk_results = Vec::new();
```
cachebag: All chunks are held in memory before writing to disk. For large files this could use significant RAM. Consider writing chunks to disk as they complete.

zb_io/src/download.rs
```diff
@@ -442,6 +495,178 @@ async fn fetch_bearer_token_internal(
...
+    let mut handles = Vec::new();
+    for chunk in chunks {
+        let client = client.clone();
+        let token_cache = token_cache.clone();
+        let url = url.to_string();
+        let total_downloaded = total_downloaded.clone();
+        let progress = progress.clone();
+        let name = name.clone();
+
+        let handle = tokio::spawn(async move {
```
cachebag, line 773: There's no retry logic here? If I understand correctly- a failed chunk just makes the entire download shit the bed.

zb_io/src/download.rs
```diff
@@ -442,6 +495,178 @@ async fn fetch_bearer_token_internal(
...
+            AUTHORIZATION,
+            HeaderValue::from_str(&format!("Bearer {token}")).unwrap(),
+        );
+    }
+
+    let response = request.send().await.map_err(|e| Error::NetworkFailure {
+        message: format!("chunk download failed: {e}"),
+    })?;
+
+    let response = if response.status() == StatusCode::UNAUTHORIZED {
```
cachebag: Nit: if a token expires in the middle of a long download, it would just fail which is not super desirable here. We should refresh the token and try again.

zb_io/src/download.rs
```diff
@@ -442,6 +495,178 @@ async fn fetch_bearer_token_internal(
...
+
+            // Report progress
+            let downloaded = total_downloaded.fetch_add(chunk_data.len() as u64, Ordering::Release);
+            if let (Some(cb), Some(n)) = (&progress, &name) {
+                cb(InstallProgress::DownloadProgress {
+                    name: n.clone(),
+                    downloaded: downloaded + chunk_data.len() as u64,
+                    total_bytes: Some(file_size),
+                });
+            }
```
cachebag: Progress is only reported after each full chunk completes (5MB). Maybe we could report progress incrementally as bytes stream in? Curious to hear your thoughts.

zb_io/src/download.rs
```diff
@@ -442,6 +495,178 @@ async fn fetch_bearer_token_internal(
...
+        response
+    };
+
+    if !response.status().is_success() {
+        return Err(Error::NetworkFailure {
+            message: format!("chunk download returned HTTP {}", response.status()),
+        });
+    }
+
+    let bytes = response.bytes().await.map_err(|e| Error::NetworkFailure {
```
cachebag: This does not validate that the server returned the correct content-range header. Parse the response header to make sure the byte range was returned.

zb_io/src/download.rs
```diff
@@ -23,6 +23,12 @@ const RACING_CONNECTIONS: usize = 4;
 /// Delay between starting each racing connection (ms)
 const RACING_STAGGER_MS: u64 = 200;
 
+/// Minimum file size to use chunked downloads (10MB)
+const CHUNKED_DOWNLOAD_THRESHOLD: u64 = 10 * 1024 * 1024;
+
+/// Size of each chunk for parallel downloads (5MB)
+const CHUNK_SIZE: u64 = 5 * 1024 * 1024;
```
cachebag: Beyond the magic numbers, I think the real issue here is that there is zero justification for the values. Why 10 MB threshold? Why 5 MB chunks? It seems arbitrary, they also do not relate to each other logically in this case particularly.

I suggest something like ...

```rust
fn calculate_chunk_size(file_size: u64) -> u64 {
    let target_chunks = 6;
    let chunk_size = file_size / target_chunks;
    chunk_size.clamp(5 * 1024 * 1024, 20 * 1024 * 1024)
}
```
This at least let's us adapt as needed..down the line it might be changed anyways but I definitely don't see this line as it is right now, a desirable addition.

fatelei:
> Looks better overall - could you ensure you make meaningful commit messages? No need for any specific format since we don't have one (as of now) but I'd suggest you rebase and reword these commits to a more thoughtful description than just "fix review comment".
> 
> We want to know what the commit did, concisely, should we ever need to revert or go back and take a look.

update

zb_io/src/download.rs
```diff
@@ -442,6 +515,298 @@ async fn fetch_bearer_token_internal(
...
+            })?;
+
+        {
+            let mut h = hasher.lock().unwrap();
+            h.update(&chunk_data);
+        }
+
+        chunks_written += 1;
+    }
+
```
cachebag: Hmm this is a bit naive. Chunks are downloaded in parallel and arrive in random order (e.g., chunk 3 might arrive before chunk 1). You're hashing chunks as they arrive, but writing them to the correct file positions.. 

So imagine..
```
-> File on disk: [chunk1][chunk2][chunk3][chunk4]... (correct order)
-> Hash computed over: [chunk3][chunk1][chunk4][chunk2]... (arrival order)
```

This will cause hash mismatches on nearly every download.

I would buffer chunks in order here. You cna use `BTreeMap<offset, chunk_data>`.

zb_io/src/download.rs
```diff
@@ -442,6 +515,298 @@ async fn fetch_bearer_token_internal(
...
+            )
+            .await?;
+
+            chunk_tx
+                .send((chunk_data, chunk.offset))
+                .map_err(|e| Error::NetworkFailure {
+                    message: format!("failed to send chunk: {e}"),
+                })?;
+
+            Ok::<(), Error>(())
```
cachebag, line 800: Spawned task errors are never observed. You need to collect and await the `JoinHandles` to catch download failures.

zb_io/src/download.rs
```diff
@@ -150,6 +171,58 @@ impl Downloader {
...
+                        (false, None)
+                    }
+                }
+                _ => (false, None),
+            }
+        };
+
+        if use_chunked {
+            if let Some(size) = file_size {
+                return download_with_chunks(
```
cachebag: Chunked downloads completely bypass the racing logic that makes the rest of zerobrew fast. The existing code races multiple connections to the same URL (hitting different CDN edges) and alternate mirrors..so sure, the fastest wins but large files (>10MB) skip this entirely.

There's a few different things you can do here, so I'll leave it up to you :)

zb_io/src/download.rs
```diff
@@ -442,6 +515,298 @@ async fn fetch_bearer_token_internal(
...
+    }
+
+    // Create output file and hasher early for streaming writes
+    let mut writer =
+        blob_cache
+            .start_write(expected_sha256)
+            .map_err(|e| Error::NetworkFailure {
+                message: format!("failed to create blob writer: {e}"),
+            })?;
+
```
cachebag: Therse no verification that `chunk_data.len()` matches what was expected for this chunk position.

zb_io/src/download.rs
```diff
@@ -23,6 +25,25 @@ const RACING_CONNECTIONS: usize = 4;
 /// Delay between starting each racing connection (ms)
 const RACING_STAGGER_MS: u64 = 200;
 
+/// Minimum file size to use chunked downloads (10MB)
+const CHUNKED_DOWNLOAD_THRESHOLD: u64 = 10 * 1024 * 1024;
+
+/// Maximum concurrent chunk downloads per file
+const MAX_CONCURRENT_CHUNKS: usize = 6;
```
maxmcd, line 41: I don't want to complicate this effort too much, but I imagine you want to think about this like a pipeline with downloads going on.

Let's say our total pipeline has parallelism of 20, then we want to chunk things up so that the queue is full most of the time, and the overhead of chunking doesn't detract from the overall performance picture.

npm+others want 20-50 concurrent connections for optimal throughput: https://www.youtube.com/watch?v=eh3VME3opnE&list=PLl386dCR5QGTElF3MbltCJupNG1lHK4Nr&index=25

So I think you want total concurrency to be configured globally and not per-download, then you can experiment with what is best, and have a fixed number of things downloading at any given time that isn't determined on a per-file basis. 

Also tests would be very useful here to benchmark what is working best and understand the overhead chunking introduces.

fatelei:
> I'm curious, is there a reason you haven't addressed https://github.com/lucasgelfond/zerobrew/pull/49/changes#r2732488395, [#49 (comment)](https://github.com/lucasgelfond/zerobrew/pull/49#discussion_r2732475412) or the good point @maxmcd made in [#49 (comment)](https://github.com/lucasgelfond/zerobrew/pull/49#discussion_r2733421267)
> 
> If you disagree, I am open to discussion, but it's not conducive to just resolve the comments and not even address the issues I outlined.

```
const GLOBAL_DOWNLOAD_CONCURRENCY: usize = 20;

  impl ParallelDownloader {
      pub fn new(blob_cache: BlobCache) -> Self {
          let semaphore = Arc::new(Semaphore::new(GLOBAL_DOWNLOAD_CONCURRENCY));
          Self { downloader: ..., semaphore, ... }
      }
  }
```

zb_io/src/download.rs
```diff
@@ -442,6 +559,314 @@ async fn fetch_bearer_token_internal(
...
+        });
+
+        handles.push(handle);
+    }
+
+    // Drop our sender so the channel closes when all tasks complete
+    drop(chunk_tx);
+
+    // Buffer chunks in order (using BTreeMap) to maintain correct hash order
+    let mut received_chunks = BTreeMap::new();
```
cachebag: All chunks are buffered in memory before writing to disk. For large files, this could cause excessive memory usage. We should be streaming writes as chunks arrive in order instead of buffering everything.

zb_io/src/download.rs
```diff
@@ -442,6 +559,314 @@ async fn fetch_bearer_token_internal(
...
+            );
+        }
+
+        match request.send().await {
+            Ok(response) => {
+                if response.status() == StatusCode::UNAUTHORIZED {
+                    match fetch_bearer_token_internal(
+                        ctx.client,
+                        ctx.token_cache,
+                        "Bearer realm=\"unknown\" scope=\"repository:homebrew/core:pull\" service=\"ghcr.io\"",
```
cachebag: Hardcoded eader in chunk downloads won't work if the server returns a different challenge. Extract the actual challenge from the response headers instead.

maxmcd:
Looks great!

fatelei:
> Thank you. LGTM, please just fix the commit messages to match the format we require in the contribution guidelines.
> 
> Appreciate all of your hard work on this!

update

cachebag (changes requested):
Image

Sorry to continue blocking this, but these commit messages don't match the changes you make at all. Could you just make them reflect the actual code changes you made instead of just "apply cargo fmt formatting", for example?

Also it's kind of ironic that `fmt` isn't passing lol

fatelei:
> Image
> Sorry to continue blocking this, but these commit messages don't match the changes you make at all. Could you just make them reflect the actual code changes you made instead of just "apply cargo fmt formatting", for example?
> 
> Also it's kind of ironic that `fmt` isn't passing lol

rebase it

cachebag (approved):
Thank you!

@maria-rcks do you have the bandwidth to test this?

cachebag (approved):
Tested myself. It looks fine, but there's a few more issues with this implementation I'll be tracking in #131

Issue #48, opened 2026-01-27
stow binary has unresolved @@HOMEBREW_PERL@@ placeholder


leeweisern:
## Description

After installing `stow` via zerobrew (`zb install stow`), the binary fails to run with:

```
zsh:1: /opt/zerobrew/prefix/bin/stow: bad interpreter: @@HOMEBREW_PERL@@: no such file or directory
```

## Cause

The shebang line in the installed binary was never interpolated:

```bash
$ head -1 /opt/zerobrew/prefix/bin/stow
#!@@HOMEBREW_PERL@@
```

It should be something like `#!/usr/bin/perl` or `#!/opt/homebrew/bin/perl`.

## Environment

- macOS Sequoia (arm64)
- zerobrew installed at `/opt/zerobrew`

## Workaround

Install via Homebrew instead: `brew install stow`

lucasgelfond:
my bad, i'm pretty sure this should be fixed as of #50  - accepted a PR (my fault) I didn't realize would break - let me know if is all good now?

leeweisern:
Thanks for the quick response! I pulled the latest changes (commit 44c9191) and rebuilt, but the issue persists.

## Reproduction

```bash
# Updated zerobrew
cd ~/.zerobrew && git pull && cargo build --release
cp target/release/zb ~/.local/bin/zb

# Cleared cache and reinstalled
zb gc
zb uninstall stow
rm -rf /opt/zerobrew/cache/
zb install stow
```

## Result

```bash
$ head -1 /opt/zerobrew/prefix/bin/stow
#!@@HOMEBREW_PERL@@

$ stow --version
zsh:1: /opt/zerobrew/prefix/bin/stow: bad interpreter: @@HOMEBREW_PERL@@: no such file or directory
```

## Environment

- macOS Sequoia 15.3 (arm64)
- zerobrew commit: 44c9191
- stow version in formula: 2.4.1

The `@@HOMEBREW_PERL@@` placeholder in the shebang line still isn't being replaced. Looking at the extract.rs changes in #50, it seems to handle `@@HOMEBREW_PREFIX@@` and `@@HOMEBREW_CELLAR@@` but perhaps `@@HOMEBREW_PERL@@` needs to be added as well?

For reference, Homebrew replaces this with the path to perl, e.g., `/opt/homebrew/opt/perl/bin/perl` or `/usr/bin/perl`.

Happy to test any fixes!

van-sprundel:
[...]

PR #51, opened 2026-01-27
(WIP) feat: Add Linux Support

hanthor:
https://github.com/user-attachments/assets/5e23e193-b58f-40cf-aaaa-f87604377fb4

This changes zerobrew to use `rustls` instead of `openssl-sys`

This adds Linux support:

- add `arwen` crate for ELF  RPATH/RUNPATH/INTERP patching
- add `elfpatch.rs` for the needed function to patch binaries
- make ZERBOBREW_ROOT settable since /opt is often a no-go on linux systems

## Try it Out

This will install the this testing linux version of zerobrew
```
 curl -sSL https://raw.githubusercontent.com/hanthor/zerobrew/installer/linux/install.sh | bash
```

### DIFF
--- a/zb_io/src/linux_patch.rs
+++ b/zb_io/src/linux_patch.rs
@@ -94,13 +94,16 @@ fn detect_zerobrew_glibc(prefix_dir: &Path) -> Option<PathBuf> {
 fn find_system_ld_so() -> Option<PathBuf> {
     // Common paths for system dynamic linkers on Linux
     let candidates = [
-        "/lib64/ld-linux-x86-64.so.2", // x86_64
-        "/lib/ld-linux-aarch64.so.1",  // aarch64/ARM64
-        "/lib/ld-linux-armhf.so.3",    // ARM hard float
-        "/lib/ld-linux.so.3",          // ARM
-        "/lib/ld-linux.so.2",          // old ARM
-        "/lib64/ld64.so.2",            // ppc64
-        "/lib64/ld64.so.1",            // s390x
+        "/lib64/ld-linux-x86-64.so.2",     // x86_64
+        "/usr/lib64/ld-linux-x86-64.so.2", // x86_64
+        "/lib/ld-linux-aarch64.so.1",      // aarch64/ARM64
+        "/usr/lib/ld-linux-aarch64.so.1",  // aarch64/ARM64
+        "/lib/ld-linux-armhf.so.3",        // ARM hard float
+        "/usr/lib/ld-linux-armhf.so.3",    // ARM hard float
+        "/lib/ld-linux.so.3",              // ARM
+        "/lib/ld-linux.so.2",              // old ARM
+        "/lib64/ld64.so.2",                // ppc64
+        "/lib64/ld64.so.1",                // s390x
     ];
 
     for candidate in &candidates {


zb_cli/src/main.rs
```diff
@@ -16,11 +16,11 @@ use zb_io::{InstallProgress, ProgressCallback};
 #[command(version)]
 struct Cli {
     /// Root directory for zerobrew data
-    #[arg(long, default_value = "/opt/zerobrew")]
+    #[arg(long, env = "ZEROBREW_ROOT", default_value = "/opt/zerobrew")]
```
hanthor: /opt shouldn't be used on linux, this a bad practice. homebrew uses ~/.linuxbrew/.linuxbrew we'll need input on where linux should default to

zb_io/src/download.rs
```diff
@@ -18,6 +18,9 @@ use crate::progress::InstallProgress;
 use zb_core::Error;
 
 /// Number of parallel connections to race when downloading (hits different CDN edges)
+#[cfg(target_os = "linux")]
+const RACING_CONNECTIONS: usize = 1;
```
hanthor, line 23: My connections stall with multiple so we set to 1

hanthor:
It works, I need guidance as to whether we should continue in the same behavior as homebrew with putting things in the linuxbrew user

```
[7/100] Benchmarking: python@3.14
Uninstalling /home/linuxbrew/.linuxbrew/Cellar/python@3.14/3.14.2_1... (3,309 files, 78.2MB)
==> Uninstalling python@3.14...
  Installing with homebrew...
  Homebrew: 21355ms
Uninstalling /home/linuxbrew/.linuxbrew/Cellar/python@3.14/3.14.2_1... (3,309 files, 78.2MB)
  Installing with zerobrew (cold)...
  Zerobrew cold: 9285ms
  Installing with zerobrew (warm)...
  Zerobrew warm: 587ms
```

zb_cli/src/main.rs
cachebag: Should either be `~/.local/share/zerobrew` or `~/.zerobrew`.. @lucasgelfond what do you think?

zb_io/src/download.rs
cachebag: This should be configurable via CLI args, not hard-coded per OS.

zb_io/src/lib.rs
```diff
@@ -7,6 +7,8 @@ pub mod extract;
 pub mod install;
 pub mod link;
 pub mod materialize;
+pub mod patchelf;
+pub use patchelf::patch_elf_placeholders;
```
cachebag: Copilot ironically made a good point here- this should not be public. We don't use it in other crates.

zb_io/src/patchelf.rs
```diff
@@ -0,0 +1,115 @@
...
+/// Uses `elb` crate to natively update RPATH.
+#[cfg(target_os = "linux")]
+pub fn patch_elf_placeholders(
+    keg_path: &Path,
+    cellar_dir: &Path,
+    _pkg_name: &str,
+    _pkg_version: &str,
+) -> Result<(), Error> {
+    // Derive prefix from cellar (cellar_dir is typically prefix/Cellar)
+    let prefix = cellar_dir.parent().unwrap_or(Path::new("/home/linuxbrew/.linuxbrew"));
```
cachebag: Use the configured prefix parameter here instead.

zb_io/src/patchelf.rs
```diff
@@ -0,0 +1,115 @@
...
+        .filter(|e| e.file_type().is_file())
+        .filter(|e| {
+            if let Ok(data) = fs::read(e.path())
+                && data.len() >= 4
+            {
+                // check for ELF magic
+                return data.starts_with(b"\x7fELF");
+            }
+            false
+        })
```
cachebag, line 154: This only patches ELF binaries. You also need to patch text files (shell scripts, Python scripts, etc.) with `@@HOMEBREW_*@@` placeholders in shebangs. 

See issues #48, #55, #57.

zb_io/src/patchelf.rs
```diff
@@ -0,0 +1,115 @@
...
+            let mut perms = metadata.permissions();
+            perms.set_mode(original_mode);
+            let _ = fs::set_permissions(path, perms);
+        }
+    });
+
+    if patch_failures.load(Ordering::Relaxed) > 0 {
+        // Just a warning for now, as not all files might be patchable
+        // and we don't want to abort installation easily.
+    }
```
cachebag: This will just confuse users when packages don't work. Warning is fine, but you need to either log it or return an error. (this is slightly on us as we don't have a robust error system in place yet)

zb_io/src/patchelf.rs
```diff
@@ -0,0 +1,115 @@
...
+            patcher.set_dynamic_tag(DynamicTag::Runpath, &*lib_path_c)?;
+            
+            // Finish writes changes back to the file
+            patcher.finish()?;
+            
+            Ok(())
+        })();
+
+        if let Err(_) = result {
+             patch_failures.fetch_add(1, Ordering::Relaxed);
```
cachebag: Add which file in why. You may use `eprintln!` for now.

zb_io/src/patchelf.rs
```diff
@@ -0,0 +1,115 @@
...
+use std::sync::atomic::{AtomicUsize, Ordering};
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+
+use elb::{DynamicTag, Elf, ElfPatcher};
+use rayon::prelude::*;
+use zb_core::Error;
+
+/// Patch @@HOMEBREW_CELLAR@@ and @@HOMEBREW_PREFIX@@ placeholders in ELF binaries.
+/// Uses `elb` crate to natively update RPATH.
```
cachebag: Nice approach. Make sure you also set `RPATH` (in addition to `RUNPATH` for compatibility with older binaries.

hanthor: Yeah I'll do that, realized this after thinking about it a while

zb_cli/src/main.rs
lucasgelfond: hmm - i'm not a linux user so have non strong opinions on directory structure. my inclination based on very little is that ~/.zerobrew is a good spot, / makes sense for something to be pretty global as a package manager, but i'm easily convinced in either direction

cachebag: Makes sense to me. Let's stick with `~/.zerobrew` @hanthor

maria-rcks: we should do `$XDG_DATA_HOME` -(if not set)-> `$HOME/.local/share` and then fall back to `$HOME`
[XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/#variables)

cachebag: ping @hanthor

maria-rcks: but then again idk why using `/opt` is bad, if its meant for "Optional software" there's also `/usr/bin/` and a whole lot more directories to put binaries in, I don't think in $HOME or local would make sense. (like if you have more than one user in  your system.)

hanthor: Yeah the main selling point for homebrew on Linux is to be able to install binaries without sudo access. Also on read-only root systems (which are quite common these days)

zb_io/src/patchelf.rs
hanthor: I was confused as it seemed function was missing for mac_os. I think Linux and mac_os can share code with this

zb_cli/src/main.rs
maria-rcks: Hm, i just checked and linuxbrew does create a user dir `/home/linuxbrew` and puts its files on `/home/linuxbrew/.linuxbrew` so it might not be that far off

hanthor: Ok I set it to `~/.zerobrew` as default for Linux but made it settable

zb_io/src/lib.rs
hanthor: https://github.com/hanthor/zerobrew/blob/linux-support/zb_io/src/materialize.rs#L7-L8

zb_io/src/patchelf.rs
hanthor: take a look

zb_io/src/patchelf.rs
hanthor: I renamed the file to `linux_patch.rs` and added the text_placeholder function.

zb_io/src/patchelf.rs
hanthor: Added.

cachebag (changes requested):
https://github.com/lucasgelfond/zerobrew/pull/51#discussion_r2731862919 and https://github.com/lucasgelfond/zerobrew/pull/51#discussion_r2731866526 still weren't addressed?

Additionally, can you make sure you read our [contributing guidelines](https://github.com/lucasgelfond/zerobrew/blob/main/CONTRIBUTING.md) and fix the commit messages?

Otherwise, these changes look good to me. Can you or anyone you know test this?

hanthor (commented):
updated based on feedback

zb_io/src/linux_patch.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        // Restore permissions
+        if is_readonly {
+            let mut perms = metadata.permissions();
+            perms.set_mode(original_mode);
+            let _ = fs::set_permissions(path, perms);
+        }
+    });
+
+    if patch_failures.load(Ordering::Relaxed) > 0 {
```
cachebag: Maybe I am missing something but I believe I made a comment on this.

https://github.com/lucasgelfond/zerobrew/pull/51#discussion_r2731862919

Could you actually address it?

zb_io/src/linux_patch.rs
```diff
@@ -0,0 +1,285 @@
...
+                perms.set_mode(original_mode);
+                fs::set_permissions(path, perms)?;
+            }
+            
+            Ok(())
+        })();
+
+         if let Err(e) = result {
+             eprintln!("Warning: Failed to patch text file at {}: {}", path.display(), e);
+             patch_failures.fetch_add(1, Ordering::Relaxed);
```
cachebag: Please address this.

https://github.com/lucasgelfond/zerobrew/pull/51#discussion_r2731866526

cachebag (changes requested):
As I mentioned, please fix the commit messages. Read the requirements for commit messags in the contribution guidelines. @hanthor

hanthor:
> [#51 (comment)](https://github.com/lucasgelfond/zerobrew/pull/51#discussion_r2731862919) and [#51 (comment)](https://github.com/lucasgelfond/zerobrew/pull/51#discussion_r2731866526) still weren't addressed?

committing now

> Additionally, can you make sure you read our [contributing guidelines](https://github.com/lucasgelfond/zerobrew/blob/main/CONTRIBUTING.md) and fix the commit messages?

I'll fix this up

> Otherwise, these changes look good to me. Can you or anyone you know test this?

I tried locally on my laptop a full benchmark but I think I ran out of memory, I'll try on my desktop in a sec

hanthor:
@cachebag take a look. I'm running the benchmark now on 58/100

zb_cli/src/main.rs
```diff
@@ -309,9 +309,27 @@ fn suggest_homebrew(formula: &str, error: &zb_core::Error) {
 }
 
 async fn run(cli: Cli) -> Result<(), zb_core::Error> {
+    let root = cli.root.unwrap_or_else(|| {
+        // 1. Check if legacy global install exists (preserve behavior)
```
cachebag: While you are at it, please remove comments like this. They provide nothing to the codebase- unless you are describing specific behavior.

cachebag (changes requested):
LGTM! Please address CI. Thanks for all your hard work on this!

zb_cli/src/main.rs
cachebag: @hanthor Do you disagree? Please don't just simply resolve the comment.

hanthor:
Still haveing issues with the racing connection when downloading multiple files

zb_cli/src/main.rs
hanthor: lol sorry I though I'd already pushed the fix

hanthor:
I'm going to keep racing off for linux as it doesn't work, I'm assuming just differences in the stack. I know you suggested adding it as a configurable, but that seems out of scope for this PR, and doesn't make sense right now as it doesn't work at all on linux.

hanthor:
I'm working on the the install.sh script

hanthor:
Working on making python work as a depedeny, I wasn't patching the interpreter parts of the ELF

Ckyzin (dismissed):

hanthor:
Ok this is working. Need to do some more tests on more computers

cachebag:
> Ok this is working. Need to do some more tests on more computers

Could you please provide screenshots, gifs, etc. some sort of verification that it actually is working?

lucasgelfond:
pls rebase also!

hanthor:
https://github.com/user-attachments/assets/5e23e193-b58f-40cf-aaaa-f87604377fb4

cachebag (approved):
This LGTM! Pinging @lucasgelfond to make sure I'm not crazy.

FYI we may wait to merge this until #74 is merged

hanthor:
Switched to the using the crate `arwen` with gobline and object as `elb` was failing to patch some stuff.

added a linux-test in a container so anyone can test linux support. https://github.com/hanthor/zerobrew/tree/linux-test/dev/linux-test

hanthor:
Need someone to verify this is non-breaking change on macos since I don't have a mac on hand

hanthor:
but adding linux support means we could add more comprehensive smoke tests in CI

zb_cli/src/main.rs
```diff
@@ -18,12 +18,12 @@ use zb_io::{InstallProgress, ProgressCallback};
 #[command(version)]
 struct Cli {
     /// Root directory for zerobrew data
-    #[arg(long, default_value = "/opt/zerobrew")]
-    root: PathBuf,
+    #[arg(long, env = "ZEROBREW_ROOT")]
+    root: Option<PathBuf>,
```
maria-rcks, line 22: In the justfile we use `ZEROBREW_OPT` as the variable name, it be nice if you could update the `Justfile` to use this too.

zb_cli/src/main.rs
```diff
@@ -359,9 +359,26 @@ async fn run(cli: Cli) -> Result<(), zb_core::Error> {
...
     }
 
+    let root = cli.root.unwrap_or_else(|| {
+        let legacy_root = PathBuf::from("/opt/zerobrew");
+        if legacy_root.exists() {
+            return legacy_root;
+        }
+        // Default to ~/.zerobrew on linux
+        if cfg!(target_os = "linux") {
+            std::env::var("HOME")
```
maria-rcks, line 369: maybe follow the [XDG Base Directory](https://specifications.freedesktop.org/basedir/latest/#variables) spec and search first for:
`XDG_DATA_HOME || HOME/.local/share/`

zb_io/src/download.rs
maria-rcks: ^^^

zb_io/src/link.rs
```diff
@@ -21,256 +22,228 @@ impl Linker {
...
     pub fn link_keg(&self, keg_path: &Path) -> Result<Vec<LinkedFile>, Error> {
-        // Create opt symlink: /opt/homebrew/opt/<name> -> /opt/homebrew/Cellar/<name>/<version>
         self.link_opt(keg_path)?;
-
-        let keg_bin = keg_path.join("bin");
-
-        if !keg_bin.exists() {
-            return Ok(Vec::new());
+        let mut linked = Vec::new();
+        for dir_name in ["bin", "lib", "include", "share"] {
```
maria-rcks, line 40: `/usr/bin/` `/usr/local/bin` `/usr/include...`

zb_io/src/link.rs
```diff
@@ -21,256 +22,228 @@ impl Linker {
...
     pub fn unlink_keg(&self, keg_path: &Path) -> Result<Vec<PathBuf>, Error> {
-        // Remove opt symlink
         self.unlink_opt(keg_path)?;
-
-        let keg_bin = keg_path.join("bin");
-
-        if !keg_bin.exists() {
-            return Ok(Vec::new());
+        let mut unlinked = Vec::new();
+        for dir_name in ["bin", "lib", "include", "share"] {
```
maria-rcks, line 127: consider `/usr/`

zb_io/src/linux_patch.rs
```diff
@@ -0,0 +1,498 @@
...
+    // Common paths for system dynamic linkers on Linux
+    let candidates = [
+        "/lib64/ld-linux-x86-64.so.2", // x86_64
+        "/lib/ld-linux-aarch64.so.1",  // aarch64/ARM64
+        "/lib/ld-linux-armhf.so.3",    // ARM hard float
+        "/lib/ld-linux.so.3",          // ARM
+        "/lib/ld-linux.so.2",          // old ARM
+        "/lib64/ld64.so.2",            // ppc64
+        "/lib64/ld64.so.1",            // s390x
+    ];
```
maria-rcks, line 107: `/usr/lib/ld-linux-x86-64.so.2`

maria-rcks (changes requested):
I'll build and test but this is what i've noted so far

zb_cli/src/main.rs
hanthor:
```suggestion
           let root = cli.root.unwrap_or_else(|| {
    let legacy_root = PathBuf::from("/opt/zerobrew");
    
    // Priority 1: Use /opt/zerobrew if the user has already set it up there
    if legacy_root.exists() {
        return legacy_root;
    }

    if cfg!(target_os = "linux") {
        // Priority 2: XDG_DATA_HOME/zerobrew
        // Priority 3: $HOME/.local/share/zerobrew
        std::env::var("XDG_DATA_HOME")
            .map(PathBuf::from)
            .ok()
            .filter(|p| !p.as_os_str().is_empty())
            .map(|p| p.join("zerobrew"))
            .unwrap_or_else(|| {
                std::env::var("HOME")
                    .map(|h| PathBuf::from(h).join(".local/share/zerobrew"))
                    .unwrap_or(legacy_root)
            })
    } else {
        // Default for macOS (non-XDG)
        legacy_root
    }
});
```

zb_io/src/link.rs
hanthor: idk this is how homebrew does it without the /usr

maria-rcks: well that could be changed easily if it's a problem, everything else LGTM

maria-rcks (approved):
LGTM

lucasgelfond (approved):
very excited about this - @maria-rcks and @cachebag have tested on linux, doesn't break for me on MacOS! merging!!

hanthor:
@maria-rcks	Were you thing thinking about this setup on Linux? 

Source: `~/.local/share/zerobrew/`
Prefix Root: `~/.zerobrew/{(bin/,Cellar/}`
Binary Executable: `~/.local/bin/zb`

PR #58, opened 2026-01-27
Clone zerobrew repository with depth 1

AfkaraLP:
clone with depth 1 so that we don't clone the entire history which is not needed for the install

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -32,7 +32,7 @@ if [[ -d "$ZEROBREW_DIR" ]]; then
     git pull
 else
     echo "Cloning zerobrew..."
-    git clone "$ZEROBREW_REPO" "$ZEROBREW_DIR"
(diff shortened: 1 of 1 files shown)


AfkaraLP:
I just encountered an error when trying to update, since we only clone the latest commit there is a mismatch in the history so we'd need to manually fetch the latest changes and then force the repo state to that version

the updated command would be 
```bash
    git fetch --depth=1 origin main
    git reset --hard origin/main
```

cachebag:
> I just encountered an error when trying to update, since we only clone the latest commit there is a mismatch in the history so we'd need to manually fetch the latest changes and then force the repo state to that version
> 
> the updated command would be
> 
> ```shell
>     git fetch --depth=1 origin main
>     git reset --hard origin/main
> ```

Could you open up a PR for this?

AfkaraLP:
on it

PR #68, opened 2026-01-27
feat(zb): Add tap support to install flow

dl-alexandre:
**Summary**
- Description: Add tap persistence and tap-aware formula resolution so installs can pull formulas from tapped repos or explicit `owner/repo/name` references.

### DIFF
--- a/zb_cli/src/utils.rs
+++ b/zb_cli/src/utils.rs
@@ -36,22 +36,6 @@ pub fn suggest_homebrew(formula: &str, error: &zb_core::Error) {
     eprintln!();
 }
 
-pub fn suggest_homebrew(formula: &str, error: &zb_core::Error) {
-    eprintln!();
-    eprintln!(
-        "{} This package can't be installed with zerobrew.",
-        style("Note:").yellow().bold()
-    );
-    eprintln!("      Error: {}", error);
-    eprintln!();
-    eprintln!("      Try installing with Homebrew instead:");
-    eprintln!(
-        "      {}",
-        style(format!("brew install {}", formula)).cyan()
-    );
-    eprintln!();
-}
-
 pub fn get_root_path(cli_root: Option<PathBuf>) -> PathBuf {
     if let Some(root) = cli_root {
         return root;


maria-rcks:
Reading through this seems like a nice thoughtful PR!  i'll test and check this later and thank you for your efforts
If you have anything to add like examples where this works and didn't before it would be great!

zb_cli/src/main.rs
```diff
@@ -308,6 +328,17 @@ fn suggest_homebrew(formula: &str, error: &zb_core::Error) {
     eprintln!();
 }
 
+fn parse_tap_name(tap: &str) -> Result<(String, String), zb_core::Error> {
```
cachebag: We prefer that this function should be moved before the `run()` function that calls it.

zb_cli/src/main.rs
```diff
@@ -323,6 +354,80 @@ async fn run(cli: Cli) -> Result<(), zb_core::Error> {
         ensure_init(&cli.root, &cli.prefix)?;
     }
 
+    if matches!(cli.command, Commands::Tap { .. } | Commands::Untap { .. }) {
```
cachebag: Wouldn't this mean users won't have access to a properly initialized API client? The current installer is created after this block, so any future tap-aware operations would be inconsistent.

zb_io/src/api.rs
```diff
@@ -32,8 +32,20 @@ impl ApiClient {
         self
     }
 
+    pub fn base_url(&self) -> &str {
```
cachebag: Why is this public? we don't use it anywhere else from what I see in your diff.

zb_io/src/install.rs
```diff
@@ -45,6 +45,73 @@ struct ProcessedPackage {
...
+    let parts: Vec<&str> = input.split('/').collect();
+    match parts.len() {
+        1 => Ok(FormulaRef {
+            name: parts[0].to_string(),
+            tap: None,
+            exact_tap: false,
+        }),
+        3 => {
+            if parts.iter().any(|p| p.is_empty()) {
+                return Err(Error::InvalidTap {
```
cachebag: This is a parsing error not a tap error, no? Shouldn't we make a new variant for this? Something like `InvalidFormulaRef`

zb_io/src/install.rs
```diff
@@ -162,54 +229,165 @@ impl Installer {
 
     /// Recursively fetch a formula and all its dependencies in parallel batches
     async fn fetch_all_formulas(&self, name: &str) -> Result<BTreeMap<String, Formula>, Error> {
```
cachebag: It's time we split the logic here into helpers.

zb_io/src/install.rs
```diff
@@ -162,54 +229,165 @@ impl Installer {
...
 
-                formulas.insert(batch[i].clone(), formula);
+                formulas.insert(name, formula);
             }
         }
 
         Ok(formulas)
     }
 
+    async fn fetch_formula_with_resolution(
```
cachebag: We silently swallow `MissingFormula` errors and try th next source...this hides useful debugging info - should accumulate and report which sources were tried.

zb_core/src/errors.rs
```diff
@@ -11,6 +11,8 @@ pub enum Error {
     MissingFormula { name: String },
     DependencyCycle { cycle: Vec<String> },
     NotInstalled { name: String },
+    InvalidTap { tap: String },
+    ConflictingFormulaSource { name: String, first: String, second: String },
```
cachebag: Both of these are missing tests.

zb_io/src/storage/db.rs
```diff
@@ -16,6 +16,14 @@ pub struct InstalledKeg {
     pub installed_at: i64,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TapRecord {
```
cachebag, line 20: No tests for this either?

dl-alexandre:
All requested items addressed:

- Added binary-only tap formula support (url+sha256+bin.install), including install path and tests.
- Updated README tap docs to reflect binary-only formulas.
- Added a tap base URL override for tests and scoped it with an env guard to avoid leakage.
- Full test suite passes (`cargo test`).

Let me know if you want any follow-ups or adjustments.

cachebag:
Best these changes begin from HEAD on main. Or resolve the conflicts accordingly.

cachebag:
Closing due to inactivity. I don't know if you're having trouble with your branch but it's probably best you just start on a fresh PR from HEAD of `main`

PR #72, opened 2026-01-27
install.sh: Support ENV overrides & refactor PATH addition

mcint:
Support overriding the install and bin paths by pre-set environment variable on first run Use .cargo/env's path addition case statement, wrapped in function, to add paths if not present (e.g. on rc re-source)

On PATH additions, this implements a standard idiom, from .cargo/env default no less, to support idempotent path addition, i.e. add once don't re-add, so users can re-source shell rc files.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,8 +5,8 @@ set -e
 # Usage: curl -sSL https://raw.githubusercontent.com/lucasgelfond/zerobrew/main/install.sh | bash
 
 ZEROBREW_REPO="https://github.com/lucasgelfond/zerobrew.git"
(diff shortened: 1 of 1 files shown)


mcint:
zomg, nice merge speed.

maria-rcks:
> ./install.sh: line 75: /home/x/.config/shell/.zshenv: Permission denied

@mcint check if you can write to the file before cat'ing

cachebag:
> zomg, nice merge speed.

Premature it looks like! The changed looked good to me but I think @maria-rcks made a good point about writing to the file first. 

Good job otherwise

Issue #54, opened 2026-01-27
Add a option to avoid modifying .zshenv


lbr77:
My ~/.zsh* files are maintained by nix-darwin so it's been set with permission r-- (so impossible to modifiy by other program unless sudo). I'think it will be better to have a more option for avoid this action.

maria-rcks:
Would it be be better if it cant modify it to tell the user add "..." to PATH in your shell config?

zcking:
I'd like this suggestion; I use [chezmoi](https://chezmoi.io/) to manage ~/.zshenv. 

I like your idea @maria-rcks 
1. attempt to modify the user's ~/.zshenv
2. if it fails, continue with initialization but print a message telling the user to add it to their PATH

rbutoi:
Just opened https://github.com/lucasgelfond/zerobrew/pull/146 which should address this

PR #73, opened 2026-01-27
Fix path export and update _zb_path_append function

mcint:
Updated the install script to persist & export ZEROBREW_DIR and ZEROBREW_BIN. Modified the _zb_path_append function to handle variable paths, as expected

Fixing previous commit

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -69,6 +69,11 @@ case "$SHELL" in
         ;;
 esac
 
+if [[ ! -w $SHELL_CONFIG ]]; then
+    echo "Error, config not writable: $SHELL_CONFIG" >&2
+    exit 1
+fi
+
 # Add to PATH in shell config if not already there
 PATHS_TO_ADD=("$ZEROBREW_BIN" "/opt/zerobrew/prefix/bin")
 if ! grep -q "^# zerobrew$" "$SHELL_CONFIG" 2>/dev/null; then


mcint:
Want to add one more message, quality of life thing.

maria-rcks:
> ./install.sh: line 75: /home/anon/.config/shell/.zshenv: Permission denied

@mcint please address this (related to #54)

mcint:
> > ./install.sh: line 75: /home/anon/.config/shell/.zshenv: Permission denied
> 
> @mcint please address this (related to #54)

I hear you. However, even as a nixpkgs user (on macOS & non-nix linux), imo, packaging for nix is a downstream concern.  I just messed up some basic bash syntax, and this and previous PR don't event have complex code paths.

_Why...how... should the tool which performs mutable install steps, serve the environment that's explicitly, and voluntarily, immutable and declaratively defined. `install.sh` is the wrong place to serve nix users. If anything the script should bail first thing._

_I could feasibly imagine attempting something like this by saving the shell_config update diff to an env var, then detecting whether to send it to stdout or append to a file._ But this is complicated, epicycles on epicycles, serving users who are well able to serve themselves, or create a mutable temporary environment, diff, or define a nix derivation.

_Requesting that a niche, self-imposed concern get solved in the core install path is how we get spaghetti software. It would be cool if nix-first specification allow generating diffs that worked for mutable systems, allowing nix to be the golden target for projects---but I don't see the nix & broader ecosystem there._

---

I'm fixing something I broke. If you're blocking it, I'm no longer the one blocking the fix.

maria-rcks:
Im not talking about NixOS support, sorry for the confusion. But we can't control we're this tool is intalled, so making sure we can write to a file seems like a nice adition to the script `[[ -w..`, or we just add it to path and let `zb init` handle it, so the code doesn't repeat itself.

install.sh
maria-rcks: consider changing this to 
```bash
            echo "_zb_path_append \"$path_entry\"" >>"$SHELL_CONFIG"
```

maria-rcks: otherwise, LGTM

mcint:
Updated to bail on `[[ ! -w ...`. Nice, `zb init` does seem a wise place to handle it. Thanks, sorry for the confusion, frustration.

cachebag:
> Updated to bail on `[[ ! -w ...`. Nice, `zb init` does seem a wise place to handle it. Thanks, sorry for the confusion, frustration.

No worries! Speaking internally, I've been sharing some strong opinions about the future of this project and if we want it to be something valuable, these conversations are really crucial to have IMO.

Not that this is a major situation or anything, but generally, I prefer contributors that provide nuance to their contributions rather than throwing 1k lines of code from claude and generating some bs PR summary.

PR #76, opened 2026-01-27
feat(download): add aria2c backend with auto-detection

lockieluke:
## Summary
Add optional aria2c download backend for faster multi-connection downloads. Auto-detected when available in PATH; disable with `ZB_DISABLE_ARIA2=1`.

## Changes
- Add `zb_io/src/aria2.rs` with aria2c RPC client and daemon management
- Refactor `download.rs` with pluggable `DownloadBackend` trait
- Extract `GhcrTokenFetcher` for reuse across backends
- CLI shows "(aria2c)" indicator when active

## aria2c features
- 4 connections per file, 8 concurrent downloads
- Resume support for interrupted downloads
- Automatic fallback to built-in downloader on failure

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,17 @@ name: lint
 on:
   push:
     branches: [ main ]
+    paths:
+      - '**/*.rs'
+      - '**/Cargo.toml'
+      - 'Cargo.lock'
+      - '.github/workflows/ci.yml'
   pull_request:
+    paths:
+      - '**/*.rs'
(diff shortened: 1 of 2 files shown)


lockieluke:
Yeah it's my first time properly "vibe coding" something with Claude Code.  I thought this might be a useful contribution to further speed up downloads and it would be quite straightforward for Claude to implement but if it really isn't viable, don't worry about it, it was literally just a shot in the dark and trying things out.

Everything seems to work but honestly it didn't seem to make it any faster.

PR #81, opened 2026-01-28
fix: export `PKG_CONFIG_PATH` for zerobrew-installed packages

### LINKED ISSUE #30: hdf5: h5cc can't configure itself
`h5cc` is used by applications as a compiler wrapper to link against hdf5, but it can't find its own headers:
```
➜  ~ which h5cc
/opt/zerobrew/prefix/bin/h5cc
➜  ~ h5cc --show
Package hdf5 was not found in the pkg-config search path.
Perhaps you should add the directory containing `hdf5.pc'
to the PKG_CONFIG_PATH environment variable
Package 'hdf5' not found
Package 'hdf5_hl' not found
pkg-config failed with arguments: --define-variable=prefix=/opt/zerobrew/prefix --cflags --libs hdf5 hdf5_hl
```

This works in Homebrew.

cachebag:
When HDF5 is installed via zerobrew, h5cc wrapper cant find `hdf5.pc` during builds. This fix sets `PKG_CONFIG_PATH` before building and exports it in user shell config for zerobrew-installed packages- which should not only fix this issue but other packages that behave like this.

Closes #30

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -39,6 +39,12 @@ fi
 
 # Build
 echo "Building zerobrew..."
+if [[ -d "/opt/zerobrew/prefix/lib/pkgconfig" ]]; then
+    export PKG_CONFIG_PATH="/opt/zerobrew/prefix/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
+fi
+if [[ -d "/opt/homebrew/lib/pkgconfig" ]] && [[ ! "$PKG_CONFIG_PATH" =~ "/opt/homebrew/lib/pkgconfig" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
dimgord, line 1: on x64_86 architecture homebrew is in /usr/local/Cellar/ and /usr/local/Caskroom/, /opt/homebrew is for arm64

uname -m -- please don't leave us behind!

cachebag: @dimgord This will come down the line. Those runners can get costly and we'd prefer to solidify what we have now before testing on intel builds.

@lucasgelfond cc

PR #83, opened 2026-01-28
fix(zb_io): formula name normalization and unsupported tap handling

quantumshiro:
issue number #67

### DIFF
--- a/zb_io/src/install.rs
+++ b/zb_io/src/install.rs
@@ -36,25 +36,6 @@ pub struct ExecuteResult {
     pub installed: usize,
 }
 
-fn normalize_formula_name(name: &str) -> Result<String, Error> {
-    let trimmed = name.trim();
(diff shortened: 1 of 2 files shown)


quantumshiro:
@cachebag plz, rereview

zb_io/src/install.rs
```diff
@@ -36,6 +36,25 @@ pub struct ExecuteResult {
     pub installed: usize,
 }
 
+fn normalize_formula_name(name: &str) -> Result<String, Error> {
```
cachebag: move this function to `zb_cli/src/main.rs` and call it before passing the name to `installer.plan()`. As it stands, the `plan()` function sits in the IO layer, but this one you added is more input validation which should be in the CLI layer.

cachebag (changes requested):
LGTM otherwise.

quantumshiro:
i fix it

PR #85, opened 2026-01-28
fix(zb_core): use revision instead of rebuild for bottle path

hanthor:
# Fix: Use Formula Revision for Bottle Installation Path

## Problem
Zerobrew was creating installation directories that included the bottle [rebuild](https://github.com/lucasgelfond/zerobrew/blob/fix/correct-bottle-installation-path/zb_core/src/formula.rs:90:4-99:5) number (e.g., `0.16.0_1`) even when the formula [revision](https://github.com/lucasgelfond/zerobrew/blob/fix/correct-bottle-installation-path/zb_core/src/formula.rs:3840:4-3842:7) was 0. This caused errors for packages like `ramalama` because the internal scripts/binaries expected the path to be the un-suffixed version (e.g., `0.16.0`).

## Solution
Updated `Formula::effective_version` in [zb_core/src/formula.rs](https://github.com/lucasgelfond/zerobrew/blob/fix/correct-bottle-installation-path/zb_core/src/formula.rs) to use the [revision](cci:1://file:///home/james/dev/zerobrew/.homebrew-reference/Library/Homebrew/formula.rb:3840:4-3842:7) field instead of `bottle.rebuild` when determining the version string for the installation directory. This ensures that:
- `revision: 0` -> `0.16.0`
- `revision: 1` -> `0.16.0_1`
- `rebuild: 1` (w/ revision 0) -> `0.16.0` (Correct behavior)

## Justification
This change aligns Zerobrew with Homebrew's internal logic, where the installation path (keg version) is derived strictly from the formula version and revision, ignoring bottle rebuilds.

**References:**
1.  **[Library/Homebrew/formula.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L642)**: The [pkg_version](cci:1://file:///home/james/dev/zerobrew/.homebrew-reference/Library/Homebrew/formula.rb:641:2-641:53) is constructed using only `version` and `revision`.
    ```ruby
    def pkg_version = PkgVersion.new(version, revision)
    ```
2.  **[Library/Homebrew/pkg_version.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/pkg_version.rb#L41-L47)**: The string representation appends `_{revision}` only if [revision](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L3841) is positive. It does not look at bottle rebuilds.
    ```ruby
    def to_str
      if revision.positive?
        "#{version}_#{revision}"
      else
        version.to_s
      end
    end
    ```

## Verification
- Verified manually by installing `ramalama` (bottle rebuild: 1).
- Installation directory is now correctly `$ZEROBREW_PREFIX/Cellar/ramalama/0.16.0`.

### DIFF
--- a/zb_core/src/bottle.rs
+++ b/zb_core/src/bottle.rs
@@ -95,6 +95,7 @@ mod tests {
             bottle: Bottle {
                 stable: BottleStable { files, rebuild: 0 },
             },
+            revision: 0,
         };
 
         let selected = select_bottle(&formula).unwrap();
@@ -123,6 +124,7 @@ mod tests {
             bottle: Bottle {
                 stable: BottleStable { files, rebuild: 0 },
             },
+            revision: 0,
         };
 
         let err = select_bottle(&formula).unwrap_err();
--- a/zb_core/src/formula.rs
+++ b/zb_core/src/formula.rs
@@ -7,15 +7,17 @@ pub struct Formula {
     pub versions: Versions,
     pub dependencies: Vec<String>,
     pub bottle: Bottle,
+    #[serde(default)]
+    pub revision: u32,
 }
 
 impl Formula {
-    /// Returns the effective version including rebuild suffix if applicable.
-    /// Homebrew bottles with rebuild > 0 have paths like `{version}_{rebuild}`.
+    /// Returns the effective version including revision suffix if applicable.
+    /// Homebrew formulas with revision > 0 have paths like `{version}_{revision}`.
+    /// Note: `rebuild` (in bottle) does NOT affect the installation directory, only the bottle filename.
     pub fn effective_version(&self) -> String {
-        let rebuild = self.bottle.stable.rebuild;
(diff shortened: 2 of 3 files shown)


cachebag (changes requested):
Thanks! Your commit message should be `fix(zb_core)` not just `core`. Please re-word and force push.

Also, please stop requesting review from co-pilot.

hanthor:
@cachebag done

zb_core/src/bottle.rs
```diff
@@ -123,6 +124,7 @@ mod tests {
             bottle: Bottle {
                 stable: BottleStable { files, rebuild: 0 },
             },
+            revision: 0,
```
cachebag, line 127: No changes needed, but eventually we'll need to reflect this in `zb_core/tests/common/mod.rs` which is a WIP PR.

Issue #24, opened 2026-01-26
Migrate from Homebrew


objectivecosta:
I know the project is still experimental (right?), but a way to migrate my formulas from brew to zb would be nice. Is this something you accept PRs on?

livcher:
Agreed, it would be a great feature. Or at least some documentation on _how_ to migrate.

I also wonder if it's at all possible to make homebrew and zerobrew coexist and sync between each other? So that if something is broken (for one package or otherwise) you can easily switch back.

ipatch:
> I also wonder if it's at all possible to make homebrew and zerobrew coexist and sync between each other? 

"this is the way"

objectivecosta:
If a PR is welcome, I think this is it: [feat(cli): add homebrew migration command #90](https://github.com/lucasgelfond/zerobrew/pull/90)

PR #90, opened 2026-01-28
feat(zb_cli/zb_io): add homebrew migration command

objectivecosta:
- Add `migrate` subcommand to import installed Homebrew formulas
- Fetch installed packages via `brew info --json` and `brew list --cask`
- Migrate CLI formulas to zerobrew and optionally uninstall from Homebrew
- Add confirmation prompts with `-y` flag for unattended migration
- Include progress reporting and failure summary

Closes https://github.com/lucasgelfond/zerobrew/issues/24

### DIFF
--- a/zb_io/src/homebrew.rs
+++ b/zb_io/src/homebrew.rs
@@ -43,23 +43,17 @@ pub fn parse_formulas_from_json(json: &serde_json::Value) -> Vec<HomebrewPackage
     packages
 }
 
-/// Parse Homebrew casks from JSON output of `brew list --cask --json=v1`
-pub fn parse_casks_from_json(json: &serde_json::Value) -> Vec<HomebrewPackage> {
-    let mut packages = Vec::new();
-
-    if let Some(casks) = json.as_array() {
-        for cask in casks {
-            if let Some(token) = cask.get("token").and_then(|t| t.as_str()) {
-                packages.push(HomebrewPackage {
-                    name: token.to_string(),
-                    tap: "homebrew/cask".to_string(),
-                    is_cask: true,
-                });
-            }
-        }
-    }
-
-    packages
+/// Parse Homebrew casks from plain text output of `brew list --cask`
+pub fn parse_casks_from_plain_text(output: &str) -> Vec<HomebrewPackage> {
+    output
+        .lines()
+        .filter(|line| !line.is_empty())
+        .map(|name| HomebrewPackage {
+            name: name.to_string(),
+            tap: "homebrew/cask".to_string(),
+            is_cask: true,
+        })
+        .collect()
 }
 
 /// Categorize Homebrew packages for migration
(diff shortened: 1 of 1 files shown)


maria-rcks:
>  error: Recipe `fmt` failed with exit code 1

Make sure you run `just build` to test out your changes, update acordingly and force push.

(This was added recently to CONTRIBUTING.md, make sure you read and understand the general guidelines.)

zb_cli/src/main.rs
cachebag, line 1: Can we write tests?

zb_cli/src/main.rs
```diff
@@ -569,6 +641,223 @@ async fn run(cli: Cli) -> Result<(), zb_core::Error> {
...
+                    });
+                }
+            };
+
+            if packages.is_empty() {
+                println!("No Homebrew packages installed.");
+                return Ok(());
+            }
+
+            let formulas: Vec<_> = packages.iter().filter(|p| !p.is_cask).collect();
```
cachebag: This should also filter by tap; ie only include formulas where `tap == "homebrew/core"`. Right now if someone has packages from other taps (like homebrew/php or custom taps or w.e.), the migration will try to install them and fail since zerobrew only supports core formulas.

zb_cli/src/main.rs
```diff
@@ -569,6 +641,223 @@ async fn run(cli: Cli) -> Result<(), zb_core::Error> {
...
+            let mut success_count = 0;
+            let mut failed: Vec<String> = Vec::new();
+
+            for pkg in &formulas {
+                print!("    {} {}...", style("○").dim(), pkg.name);
+
+                match installer.plan(&pkg.name).await {
+                    Ok(plan) => {
+                        // Execute the plan without progress bars for batch migration
+                        match installer.execute(plan, true).await {
```
cachebag, line 692: Having no tests for this feature is not desirable at all.

zb_cli/src/main.rs
```diff
@@ -569,6 +641,223 @@ async fn run(cli: Cli) -> Result<(), zb_core::Error> {
...
+            for pkg in &formulas {
+                // Skip if it failed to install in zerobrew
+                if failed.contains(&pkg.name) {
+                    continue;
+                }
+
+                print!("    {} {}...", style("○").dim(), pkg.name);
+
+                let status = Command::new("brew")
+                    .args(["uninstall", "--force", &pkg.name])
```
cachebag: The `--force` flag removes all installed versions. Should either document this in the confirmation prompt or make it configurable. Users might want to keep certain versions.

zb_cli/src/main.rs
lucasgelfond: +1

zb_cli/src/main.rs
lucasgelfond: +1 

I am generally pro something ike this, but I also would love an API that, for example, outputs whatever was not able to be resolved. one of the guiding design choices here was to be able to interoperate w/ homebrew so that the long tail of cases we don't (yet) support are still fine.

zb_cli/src/main.rs
lucasgelfond: I agree!!! tests for sure

zb_cli/src/main.rs
lucasgelfond: +1

objectivecosta: TIL

zb_io/src/homebrew.rs
```diff
@@ -0,0 +1,319 @@
...
+pub fn parse_casks_from_json(json: &serde_json::Value) -> Vec<HomebrewPackage> {
+    let mut packages = Vec::new();
+
+    if let Some(casks) = json.as_array() {
+        for cask in casks {
+            if let Some(token) = cask.get("token").and_then(|t| t.as_str()) {
+                packages.push(HomebrewPackage {
+                    name: token.to_string(),
+                    tap: "homebrew/cask".to_string(),
+                    is_cask: true,
```
objectivecosta: I think these two are kind of sus but I couldn't think of a better way (having a tap to denote cask and the `is_cask`)

zb_cli/src/main.rs
objectivecosta: Done! Moved most of the logic to zb_io, added tests there! There's still some logic in zb_cli, but I think it's the bare minimum for the CLI

zb_cli/src/main.rs
objectivecosta: Addressed :)

zb_cli/src/main.rs
objectivecosta: Addressed by adding a force command separately on the migrate command!

zb_io/src/homebrew.rs
cachebag: Yeahhh it's not ideal but for now it's sufficient.

cachebag (approved):
LGTM

PR #96, opened 2026-01-29
feat(zb_cli): shell compeletions for zb

Xiao-Chenguang:
Shell completions is a great feature to improve use experiences. This PR implement it for zb in zsh, bash, and fish.
Internally, clap_complete is used to support this feature. The code is generated by Copilot and double checked mannually.

Tested in zsh on a mac. Document is updated accordingly.


README.md
cachebag, line 1: We can remove this. We don't need it in the readme but eventually, we will have docs where this info might be of value.

Xiao-Chenguang:
> Thanks for the AI disclosure! I actually really like this addition.
> 
> Could you test it locally and send a gif/short clip?

Here you go.
zb_shell_completion

README.md
Xiao-Chenguang: I can remove and save it locally for now.

PR #98, opened 2026-01-29
chore: remove `Justfile` ft specify

BBboy01:
`Just` have [syntax highlighting](https://github.com/casey/tree-sitter-just) and [lsp](https://github.com/terror/just-lsp), we may doesn't need to force specify this file type

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,5 +1,3 @@
-# vim: set ft=make :
-
 set export
 set dotenv-load
 set script-interpreter := ['bash', '-euo', 'pipefail']


maria-rcks:
`just` recommends [to do this](https://just.systems/man/en/vim-and-neovim.html) and it only affects vim-type editors

BBboy01:
> `just` recommends [to do this](https://just.systems/man/en/vim-and-neovim.html) and it only affects vim-type editors

It affects neovim users too, but neovim treesitter work perfect, so I prefer this to be a personal config

cachebag (commented):
Keeping it does not affect non-Vim editors and seems to improve compatibility for Vim/Neovim users who don’t have vim-just or Tree-sitter configured. 

Seems like a useless PR in this state, if you don't want it you're free to adjust your local config when developing.

BBboy01:
> Keeping it does not affect non-Vim editors and seems to improve compatibility for Vim/Neovim users who don’t have vim-just or Tree-sitter configured.

You are right, but it force users who well configured vim-just or treesitter to use the compatible version

With ft specify but well configured neovim:
CleanShot 2026-01-30 at 11 57 57@2x

Without ft specify but well configured neovim:

CleanShot 2026-01-30 at 11 57 03@2x

BBboy01:
Since this specified ft to `make`, `just-lsp` not working too, cause the lsp is working `just` file not `make` file

cachebag:
Ping? @BBboy01

BBboy01:
It's done. Sorry, I don't often open my computer on weekends

maria-rcks (approved):
LGTM

PR #105, opened 2026-01-29
feat(zb_cli): Add install support for termux on android

abdalmoniem:
in termux, the `/opt/zerobrew` and `/opt/zerobrew/prefix` directories won't be created due to insufficient permissions enforced by the Android OS.

this PR adds the support for termux by making the default values of `Cli.root` and `Cli.prefix` configurable at compile time using `option_var`

if the project is built like:

```bash
ZB_ROOT="$PREFIX/opt/zerobrew" ZB_PREFIX="$PREFIX/opt/zerobrew/prefix" cargo build --release
```

then `Cli.root` will be set to `ZB_ROOT` and `Cli.prefix` will be set to `ZB_PREFIX`, if these environment variables are not set, then the original default values will be used.

a new `install-termux.sh` script has been created to setup the correct environment in a termux session and use the new way of building the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,13 @@ cargo build --release
 cargo install --path zb_cli
 ```
 
+or the following if you want to install on termux on android:
+
+```bash
+ZB_ROOT="$PREFIX/opt/zerobrew" ZB_PREFIX="$PREFIX/opt/zerobrew/prefix" cargo build --release
+cargo install --path zb_cli
+```
+
 ## Benchmarking
 
 ```bash


zb_cli/src/main.rs
```diff
@@ -33,6 +34,27 @@ struct Cli {
...
+    /// get the default root directory for zerobrew data
+    /// using the `ZB_ROOT` env var.
+    ///
+    /// if `ZB_ROOT` is not set, `/opt/zerobrew` is used
+    fn get_default_root() -> OsStr {
+        // Looks for ZB_ROOT at compile time; defaults to /opt/zerobrew
+        option_env!("ZB_ROOT").unwrap_or("/opt/zerobrew").into()
+    }
+
+    /// get the default prefix directory for linked binaries
```
cachebag:
```suggestion
    /// Get the default prefix directory for linked binaries
```

Veryyyy big nitpick lol

zb_cli/src/main.rs
```diff
@@ -33,6 +34,27 @@ struct Cli {
     command: Commands,
 }
 
+impl Cli {
+    /// get the default root directory for zerobrew data
```
cachebag:
```suggestion
    /// Get the default root directory for zerobrew data
```

install-termux.sh
```diff
@@ -0,0 +1,140 @@
...
+# zerobrew installer for termux on android
+# Usage: curl -sSL https://raw.githubusercontent.com/lucasgelfond/zerobrew/main/install_termux.sh | bash
+
+ZEROBREW_REPO="https://github.com/lucasgelfond/zerobrew.git"
+: ${ZEROBREW_DIR:=$HOME/.zerobrew}
+: ${ZEROBREW_BIN:=$HOME/.local/bin}
+
+echo "Installing zerobrew..."
+
+# # Check for Rust/Cargo
```
cachebag: If we aren't using the commented out section, best that we remove them.

install-termux.sh
cachebag, line 1: Not sure what the purpose is for simply duplicating the install script and tweaking it to make it work for termux. Are the changes not contained enough to be added into the `install.sh` script we already have?

cachebag (changes requested):
Thanks! Note: these should be 3 commits. i.e. changes to `zb_cli` go in `feat(zb_cli):..` install  script would be in `feat(install):...` etc. Same with updates to documentation.

install-termux.sh
abdalmoniem: that's a mistake, I commented this section for while testing and forgot to uncomment it before pushing 😅 will follow up with a new commit with this section included

install-termux.sh
abdalmoniem: I pushed new commits with the requested changes and merged `install-termux.sh` with `install.sh` please take a look at it and review it, I might've missed something 😅

cachebag (changes requested):
@abdalmoniem  Please resolve conflicts. https://github.com/lucasgelfond/zerobrew/pull/124 collapsed the zb_cli module into delegated modules so please apply your changes accordingly :?

abdalmoniem:
I think https://github.com/lucasgelfond/zerobrew/pull/51 already implemented this, now `zerobrew` can be installed on Termux, although no binaries downloaded from `zerobrew` will work since homebrew bottles are built for standard GNU/Linux distributions (like Ubuntu), not for Termux's unique Android environment. The key problems are:

1. **`Interpreter path`**: The binary expects `@@HOMEBREW_PREFIX@@/lib/ld.so` (a placeholder that should be replaced), but Termux uses `/system/bin/linker64`
2. **`Library paths`**: The binary expects `GNU libc (libc.so.6)`, but Termux uses `Android's Bionic libc`
3. **`Library dependencies`**: Homebrew bottles expect libraries in standard Linux paths, but Termux has its own prefix
4. **`The fundamental issue`**: Homebrew Linux bottles are incompatible with Termux. Termux-specific bottles are needed (there're none available as far as I know).

currently trying to install a formula will result in:
Screenshot_20260131_123647_Termux
inconvinient but at least nothing broken gets installed!

anyhow, I think this PR should be closed since it's not relevant anymore 😊

PR #117, opened 2026-01-30
feat: Taps and Formulae

hanthor:
## Description
This PR introduces support for installing formulas from third-party taps and implements a Ruby formula parser using `tree-sitter-ruby`.

### Features
- **Tap Management**: New `tap` and `untap` commands to manage third-party repositories.
- **Ruby Formula Parser**: Replaces regex-based parsing with a `tree-sitter` implementation for robust DSL support.
- **Dependency Resolution**: Correctly resolves dependencies across different taps (e.g., `kcl` -> `kcl-lsp`).
- **Mixed Content Support**: Can install from a mix of local taps and the Homebrew API.

### Verified Formulas
- `kcl` (kcl-lang/tap) - Verified dependency resolution
- `ytt` (carvel-dev/carvel) - Verified complex `if/elsif` block parsing
- `ah` (artifacthub/cmd) - Verified conditional logic
- `dapr-cli` (dapr/tap) - Verified basic tap support

### Limitations
- **Source-Only Formulas**: Formulas that require compilation (like `tiup`) are not supported as Zerobrew currently only supports binary bottles.
- **Complex Interpolation**: Formulas using complex Ruby string interpolation in URLs (like `telepresence-oss`) may fail parsing.

### Future Potential
To fully support the entire Homebrew ecosystem, including complex Ruby logic and source builds, we could integrate a full Ruby VM context. Potential approaches include:
- **Artichoke Ruby**: Embedding Artichoke for a lightweight, Rust-native Ruby implementation.
- **FFI**: Using FFI to bind to a standard CRuby VM.

### DIFF
--- a/zb_io/src/formula_parser.rs
+++ b/zb_io/src/formula_parser.rs
@@ -200,8 +200,10 @@ fn process_conditional(node: Node, source: &str, data: &mut Formula, context: &m
         if kind == "if" || kind == "elsif" {
             if let Some(condition) = n.child_by_field_name("condition") {
                 if should_process_condition_node(&condition, source) {
+                    let mut new_ctx = context.clone();
+                    update_context_from_condition(&mut new_ctx, &condition, source);
                     if let Some(body) = n.child_by_field_name("consequence") {
-                        process_branch_body(body, source, data, context);
+                        process_branch_body(body, source, data, &mut new_ctx);
                     }
                     return; // Handled
                 }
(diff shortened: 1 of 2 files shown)


hanthor:
looks like tap formulas that don't have bottles will need ruby DSL to read and execute formulae. So think I'm cooking something up that can make that work and I can bring it in here

cachebag:
Re-open if you plan on completing the work.

hanthor:
It was finished

cachebag:
> It was finished

There's several conflicts and CI is not passing. I assumed it was abandoned.

hanthor:
I'll bump it and lint it. I have few more PRs after this one

PR #123, opened 2026-01-30
fix(zb_io): isolate HTTP/2 connections per racing download to prevent stream contention

### LINKED ISSUE #116: racing connections workaround
#51 broke macOS builds 

Instead of disabling racing (as we've done in #115), the fix should isolate or eliminate HTTP/2 stream contention e.g., by using separate TLS/HTTP clients per racing connection or forcing HTTP/1.1 for raced downloads.

cachebag:
Fixes #116 

Racing connections shared a single HTTP/2 connection, bad news bears. We can avoid stream contention and flow-control blocking by sharing a TLS config and creating separate `reqwest` clients per racing connection. 

This re-enables them for both macOS and linux (woo). Tested on both macOS and Linux.

I also bumped all version deps and fixed a strangling clippy issue (bb069642276f31018188d8090dc5b3c0ae74f257) that somehow wasn't caught in CI(?)

### DIFF
--- a/zb_io/src/download.rs
+++ b/zb_io/src/download.rs
@@ -19,11 +19,7 @@ use crate::blob::BlobCache;
 use crate::progress::InstallProgress;
 use zb_core::Error;
 
-/// Number of parallel connections to race when downloading (hits different CDN edges)
-/// Note: Racing is disabled on Linux and macOS due to HTTP/2 stream contention issues with rustls
-const RACING_CONNECTIONS: usize = 1;
-
-/// Delay between starting each racing connection (ms)
+const RACING_CONNECTIONS: usize = 3;
 const RACING_STAGGER_MS: u64 = 200;
 
 /// Minimum file size to use chunked downloads (10MB)
@@ -76,6 +72,7 @@ struct ChunkedDownloadContext<'a> {
     file_size: u64,
     global_semaphore: &'a Arc<Semaphore>,
 }
+// FIXME: extract timeout and HTTP/2 window size constants to config file
 
 /// Callback for download progress updates
 pub type DownloadProgressCallback = Arc<dyn Fn(InstallProgress) + Send + Sync>;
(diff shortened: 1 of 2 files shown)


zb_io/src/download.rs
```diff
@@ -75,36 +73,70 @@ struct CachedToken {
...
+
         Self {
             client: reqwest::Client::builder()
                 .user_agent("zerobrew/0.1")
                 .pool_max_idle_per_host(10)
                 .tcp_nodelay(true)
                 .tcp_keepalive(Duration::from_secs(60))
                 .connect_timeout(Duration::from_secs(30))
-                .timeout(Duration::from_secs(300)) // 5 min total request timeout
+                .timeout(Duration::from_secs(300))
```
lucasgelfond, line 170: sidenote, fine to merge, should we put these in a config somewhere eventually?

zb_io/src/download.rs
```diff
@@ -75,36 +73,70 @@ struct CachedToken {
...
+            .user_agent("zerobrew/0.1")
+            .use_preconfigured_tls(self.tls_config.clone())
+            .pool_max_idle_per_host(0)
+            .tcp_nodelay(true)
+            .tcp_keepalive(Duration::from_secs(60))
+            .connect_timeout(Duration::from_secs(30))
+            .timeout(Duration::from_secs(300))
+            .http2_adaptive_window(true)
+            .http2_initial_stream_window_size(Some(2 * 1024 * 1024))
+            .http2_initial_connection_window_size(Some(4 * 1024 * 1024))
```
lucasgelfond, line 195: i.e. we're reusing a bunch of these args

lucasgelfond (approved):
some nits, but this looks good to me!

zb_io/src/download.rs
cachebag: I'll add a TODO/FIXME. I think right now its fine, premature optimization is the root of all evil IMO

cachebag (commented):
Perf gains are still to be seen, but comparing `master:HEAD` to this PR, we still see racing connections working (both on macOS and linux) which I guess was the initial goal of this PR.

Tracking to improve perf efforts -> #132

PR #135, opened 2026-01-31
feat(zb): `zbx` command

### LINKED ISSUE #120: `zbx`
It would be great if there was a `zbx` command, similar to `uvx`, which would allow to run any program without having to install it.

cachebag:
Introduces `zbx` command, effectively allowing us run any program without having to install it. I've also collapsed `zb_cli` to store our binaries in `bin/`.

Closes #120

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -37,7 +37,7 @@ default:
 [doc('Build the zb binary')]
 [group('build')]
 build: fmt lint
-    cargo build --bin zb
+    cargo build --bin zb --bin zbx
 
 [doc('Install zb to $ZEROBREW_BIN')]
 [group('install')]
@@ -52,7 +52,9 @@ install: build
 
     mkdir -p "$ZEROBREW_BIN"
     install -Dm755 target/debug/zb "$ZEROBREW_BIN/zb"
+    install -Dm755 target/debug/zbx "$ZEROBREW_BIN/zbx"
     echo "Installed zb to $ZEROBREW_BIN/zb"
+    echo "Installed zbx to $ZEROBREW_BIN/zbx"
 
     "$ZEROBREW_BIN/zb" init
 


zb_cli/src/bin/zbx.rs
```diff
@@ -0,0 +1,48 @@
+use console::style;
+use std::env;
+use std::process::Command;
+
+fn main() {
+    let args: Vec<String> = env::args().skip(1).collect();
+
+    if args.is_empty() {
+        eprintln!("zbx - Run a Homebrew package without installing it");
```
lucasgelfond: run a package? explicit mention of homebrew?

zb_cli/src/commands/run.rs
```diff
@@ -0,0 +1,305 @@
...
+    let normalized = normalize_formula_name(formula)?;
+
+    let was_installed = installer.is_installed(&normalized);
+
+    if !was_installed {
+        println!(
+            "{} Installing {} temporarily...",
+            style("==>").cyan().bold(),
+            style(&normalized).green()
+        );
```
lucasgelfond, line 24: one thought, we have this logic in a ton of palces, dones't need to be in this PR but we maybe want to refactor

lucasgelfond (approved):
my goat - trusting you've tested, this is awesome

zb_cli/src/bin/zbx.rs
cachebag: hmm yeah I'll just change it to 'run a command'? wdyt

zb_cli/src/commands/run.rs
cachebag: what are you referring to in specific? the whole block or the style callsites?

maria-rcks: I think he's refering to all the print's we have with the same kind of format, we have a  println wrapper as a TODO internally tho, wouldn't need to block this pr for it.

cachebag:
image

Lmk if I should run any more commands to be thorough 

cc @felipecrs

zb_cli/src/commands/run.rs
cachebag: in that case, might be worth it to look into [log](https://docs.rs/log/latest/log/)

lucasgelfond:
also the README!

PR #136, opened 2026-01-31
Termux Error Handling

abdalmoniem:
it's again with termux 😅 this time cf9cfa0de08c286749d0468bbd92db5c45024ee7 introduced a runtime panic on termux/android environments due to the `reqwest` dependency using the `rustls` which doesn't work in this environment:

Screenshot_20260131_231356_Termux
Screenshot_20260131_231425_Termux
Screenshot_20260131_231440_Termux

this PR does the following:

1. fixes this by using the `native-tls` instead of `rustls` on android configurations.
2. also added a more informative error message that indicates that it's not feasible to install the erroneous formula on this environment instead of suggesting to use `try installing using: brew install <erroneous_formaula>` which would not help and would send users down a rabbit hole 😅
Screenshot_20260131_231502_Termux
3. added a convinient `uninstall.sh` that completely uninstalls `zerobrew` instead of manually hunting down and removing `zerobrew` related files and directories.


cachebag (commented):
Will TAL I get to my desk- but how could that commit have caused this? Lol it's literally just a readme change?

I'm assuming you meant #123

abdalmoniem:
oh yeah, my mistake 😅😅😅

Cargo.toml
```diff
@@ -13,7 +13,7 @@ clap = { version = "4", features = ["derive", "env"] }
 clap_complete = "4"
 console = "0.16.2"
 indicatif = "0.18.3"
-reqwest = { version = "0.13.1", default-features = false, features = ["json", "stream", "rustls", "http2"] }
+reqwest = { version = "0.13.1", default-features = false, features = ["json", "stream", "http2"] }
```
cachebag: I'm not too sure about this. The workspace definition still has `rustls` for non-Android, but you removed it from the workspace level. This means ALL platforms lose `rustls` initially, then `zb_io` gets it back conditionally.

Could you do something like this instead?

```toml
# In Cargo.toml (workspace) 
reqwest = { version = "0.13.1", default-features = false, features = ["json", "stream", "rustls", "http2"] }

# In zb_io/Cargo.toml - override it for Android only like so...
[target.'cfg(not(target_os = "android"))'.dependencies]
reqwest = { workspace = true }

[target.'cfg(target_os = "android")'.dependencies]
reqwest = { version = "0.13.1", default-features = false, features = ["json", "stream", "native-tls", "http2"] }
```

zb_cli/src/utils.rs
```diff
@@ -28,11 +28,29 @@ pub fn suggest_homebrew(formula: &str, error: &zb_core::Error) {
...
+        eprintln!(
+            "      {} {}",
+            style(formula).yellow().bold(),
+            style("is not compatible with termux on android")
+                .red()
+                .bold()
+        );
+        eprintln!(
+            "      {}",
+            style("and can not be installed on it.").red().bold()
```
cachebag:
```suggestion
            style("and cannot be installed on it.").red().bold()
```
Major nit

zb_cli/src/utils.rs
```diff
@@ -28,11 +28,29 @@ pub fn suggest_homebrew(formula: &str, error: &zb_core::Error) {
...
-        style(format!("brew install {}", formula)).cyan()
-    );
+
+    // Error for termux on android since homebrew
+    // doesn't support bottles for this platform
+    if cfg!(target_os = "android") {
+        eprintln!(
+            "      {} {}",
+            style(formula).yellow().bold(),
+            style("is not compatible with termux on android")
```
cachebag:
```suggestion
            style("is not compatible with termux- homebrew bottles are not available for Android.")
```

We should explain why it's not compatible

uninstall.sh
cachebag, line 1: This is a bit out of scope from the PR..users can just do
```bash
zb reset --yes
rm ~/.local/bin/zb
```
This should just be removed for now or open a separate PR for it.

zb_cli/src/utils.rs
```diff
@@ -28,11 +28,29 @@ pub fn suggest_homebrew(formula: &str, error: &zb_core::Error) {
...
     eprintln!("      Error: {}", error);
     eprintln!();
-    eprintln!("      Try installing with Homebrew instead:");
-    eprintln!(
-        "      {}",
-        style(format!("brew install {}", formula)).cyan()
-    );
+
+    // Error for termux on android since homebrew
+    // doesn't support bottles for this platform
```
cachebag, line 33: Can you add a comment below this with a link to this PR so people understand why we're including this?

zb_io/Cargo.toml
```diff
@@ -28,6 +27,12 @@ zb_core = { path = "../zb_core" }
 arwen = "0.0.5"
 object = "0.38.1"
 
+[target.'cfg(not(target_os = "android"))'.dependencies]
```
cachebag, line 33: `native-tls` is so heavy. Although I like the idea for supporting Android, I think it's best we make this opt-in i.e. something like...
```toml
[features]
android-support = ["reqwest/native-tls"]
```

cachebag (changes requested):
I love this PR. I think it's a very nice addition for user compatibility.

Have you tested this fix on actual Termux/Android hardware? Mainly concerned about these here
1. The app now runs without the `rustls` panic
2. Downloads work correctly with `native-tls`
3. HTTPS connections to ghcr.io/homebrew work

Would be good to get confirmation from you or someone else before merging this.

Cargo.toml
abdalmoniem: good idea, will fix in the upcoming commit.

uninstall.sh
abdalmoniem: sure, good point

abdalmoniem:
> I love this PR. I think it's a very nice addition for user compatibility.
> 
> Have you tested this fix on actual Termux/Android hardware? Mainly concerned about these here
> 
> 1. The app now runs without the `rustls` panic
> 2. Downloads work correctly with `native-tls`
> 3. HTTPS connections to ghcr.io/homebrew work
> 
> Would be good to get confirmation from you or someone else before merging this.

these commits have been developed, tested and pushed from Termux 🤣
but to address your concerns:
1. The app now runs without the `rustls` panic
   - yes, although I think it's working fine now because there's no feasible way for the android build to call any `rustls` related stuff since the app errors out before doing any actual downloads.
2. Downloads work correctly with `native-tls`
   - downloads don't work for now since the app errors out when it doesn't find any formulas/bottles compatible with android
4. HTTPS connections to ghcr.io/homebrew work
   - I think not due to previous reasons, I haven't fully worked out how the app works yet, but I think the only connections made on android are those that check compatibility with the current platform, but then errors out and exits before doing any downloads (although I've tried to allow `arm64 linux` bottles to be downloaded on android builds and it worked fine, i.e. `zerobrew` was able to download and install formulas successfully, but those formulas didn't actually work because of the incompatibility issues mentioned in [#105](https://github.com/lucasgelfond/zerobrew/pull/105))

cachebag:
> these commits have been developed, tested and pushed from Termux 🤣

Fair enough lol I still think it would be nice to have someone else test or do so on a separate machine. But if you can't that's fine.

> downloads don't work for now since the app errors out when it doesn't find any formulas/bottles compatible with android

This is my my push for why this should be opt-in. I don't want code in here that won't be used by anyone, esp if it doesn't work right now? So the feature attribute is probably a dealbreaker here.

Also, sorry for the delay on this I don't have an excuse for why I took so long. Really appreciate your patience, and although it's not by any means a hotspot for now, I love this PR for the sole fact that it covers a use case that 99% of people wouldn't have thought about. I love this type of work!

abdalmoniem:
> Also, sorry for the delay on this I don't have an excuse for why I took so long. Really appreciate your patience, and although it's not by any means a hotspot for now, I love this PR for the sole fact that it covers a use case that 99% of people wouldn't have thought about. I love this type of work!

no worries — that’s the beauty of open source, and what makes them great 😉

cachebag (commented):
Not sure if you're done yet but I don't see any of the files updated?

abdalmoniem:
> Not sure if you're done yet but I don't see any of the files updated?

sorry, I was occupied by #183. some tests were failing so I had to investigate 😅

cachebag:
> > Not sure if you're done yet but I don't see any of the files updated?
> 
> sorry, I was occupied by #183. some tests were failing so I had to investigate 😅

No worries! I left a comment and closed that PR. You should include all changes in here since the fixes for the failing tests relate to this PRs changes.

abdalmoniem:
@cachebag please check the new commits 😊

cachebag (approved):
Thank you!

Issue #138, opened 2026-01-31
tests for multiple formulas


cachebag:
#129 doesn't have any tests

PR #129, opened 2026-01-31
feat(zb_cli): implement simultaneous multi-formula installation support

himudigonda:
This PR enables Zerobrew to accept and process multiple formulas in a single command (e.g., `zb install git node python`). It significantly improves user efficiency by parallelizing dependency resolution and fetching across the entire requested set.

## Changes
I have adapted the implementation to align with the recent modular refactor of the CLI while maintaining atomic commit standards. Solves conflicts in #106 

### 1. Core Logic (`zb_core`)
* **Refactor:** Updated `resolve_closure` to accept a slice of root formulas (`&[String]`).
* **Logic:** The dependency graph now calculates a unified transitive closure for all requested packages, ensuring shared dependencies are only resolved once.

### 2. IO & Installer (`zb_io`)
* **Feat:** Updated `Installer::plan` and `Installer::install` to accept multiple formula names.
* **Optimization:** Implemented `fetch_all_formulas` using `HashSet` for deduplication and parallelized API requests to minimize network overhead.

### 3. Modular CLI (`zb_cli`)
* **Schema:** Updated `Commands::Install` in `cli.rs` to use `Vec<String>` with `num_args = 1..`.
* **Execution:** Refactored `commands/install.rs` to handle the vector of formulas and correctly manage multiple progress bars via `indicatif`.
* **Routing:** Updated the main router in `main.rs` to delegate the new plural `formulas` argument to the install command handler.

PR #139, opened 2026-02-01
test(zb_cli): add integration tests for {,un}install, list, info, gc 

### LINKED ISSUE #119: full integration tests
would love to see PRs for this. a lot of how i'd test in initial dev was just installing a series of packages and making sure they still worked, ie

zb install ffmpeg
ffmpeg

maybe make a gif from an mp4, etc

there's probably a stronger methodology here, but we are missing certain breaks because the rust unit tests only cover so much surface area

rbutoi:
Resolves #119. These tests are intentionally minimal / not validating extensively but they should detect regressions.

### DIFF
--- a/zb_cli/tests/integration.rs
+++ b/zb_cli/tests/integration.rs
@@ -1,0 +1,212 @@
+use std::path::PathBuf;
+use std::process::{Command, Output};
+
+struct TestEnv {
+    root: tempfile::TempDir,
+}
+
+impl TestEnv {
+    fn new() -> Self {
+        Self {
+            root: tempfile::TempDir::new().expect("failed to create temp dir"),
+        }
+    }
+
+    fn zb(&self, args: &[&str]) -> Output {
+        let zb = env!("CARGO_BIN_EXE_zb");
+        Command::new(zb)
+            .env("ZEROBREW_ROOT", self.root.path())
+            .args(args)
+            .output()
+            .unwrap_or_else(|_| panic!("failed to execute {zb} command"))
+    }
+
+    fn bin_dir(&self) -> PathBuf {
+        self.root.path().join("prefix").join("bin")
+    }
+
+    fn count_store_entries(&self) -> usize {
+        assert!(self.root.path().join("store").is_dir());
+        std::fs::read_dir(self.root.path().join("store"))
+            .map(|r| r.count())
+            .expect("failed to read store directory")
+    }
+
+    fn run_binary(&self, name: &str, args: &[&str]) -> Output {
+        let bin_path = self.bin_dir().join(name);
+        Command::new(&bin_path)
+            .env(
+                "PATH",
+                format!(
+                    "{}:{}",
+                    self.bin_dir().display(),
+                    std::env::var("PATH").unwrap_or_default()
+                ),
+            )
+            .args(args)
+            .output()
+            .unwrap_or_else(|e| panic!("failed to execute {}: {e}", bin_path.display()))
+    }
+}
+
+fn assert_success(output: &Output, context: &str) {
+    assert!(
+        output.status.success(),
+        "{} failed:\nstdout: {}\nstderr: {}",
+        context,
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr)
+    );
+}
+
+fn assert_stdout_contains(output: &Output, needle: &str) {
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(
+        stdout.contains(needle),
+        "expected stdout to contain {needle:?}, got: {stdout}"
+    );
+}
+
+fn assert_no_installed_symlinks(dir: &std::path::Path) {
+    if !dir.exists() {
+        return;
+    }
+    let cellar = dir.join("Cellar");
+    for entry in walkdir::WalkDir::new(dir) {
+        let entry = entry.expect("failed to read directory entry");
+        if entry.path().starts_with(&cellar) {
+            continue;
+        }
+        assert!(
+            !entry.path_is_symlink(),
+            "unexpected symlink: {}",
+            entry.path().display()
+        );
+    }
+}
+
+#[test]
+#[ignore = "integration test"]
+#[cfg(target_os = "macos")] // GitHub Actions linux runner needs additional X11/XCB deps
+fn test_ffmpeg_formula() {
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["install", "ffmpeg"]), "zb install ffmpeg");
+
+    // From the upstream formula test:
+    // https://github.com/Homebrew/homebrew-core/blob/3076627c980d101ff02a720060c508433c44f293/Formula/f/ffmpeg.rb#L114
+    let mp4out = t.root.path().join("video.mp4");
+    assert_success(
+        &t.run_binary(
+            "ffmpeg",
+            &[
+                "-filter_complex",
+                "testsrc=rate=1:duration=5",
+                mp4out.to_str().unwrap(),
+            ],
+        ),
+        "ffmpeg create test video",
+    );
+    assert!(mp4out.exists());
+}
+
+#[test]
+#[ignore = "integration test"]
+fn test_curl_simple() {
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["install", "curl"]), "zb install curl");
+
+    let output = t.run_binary("curl", &["https://www.githubstatus.com"]);
+    assert_success(&output, "curl https://www.githubstatus.com");
+    assert_stdout_contains(&output, "GitHub");
+}
+
+#[test]
+#[ignore = "integration test"]
+fn test_install_uninstall_and_reinstall() {
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["install", "jq"]), "zb install jq");
+
+    let test_json = t.root.path().join("test.json");
+    std::fs::write(&test_json, r#"{"foo":1, "bar":2}"#).expect("failed to write test.json");
+
+    let output = t.run_binary("jq", &[".bar", test_json.to_str().unwrap()]);
+    assert_success(&output, "jq .bar test.json");
+    assert_eq!(String::from_utf8_lossy(&output.stdout), "2\n");
+
+    assert_success(&t.zb(&["uninstall", "jq"]), "zb uninstall jq");
+    assert_success(&t.zb(&["uninstall", "oniguruma"]), "zb uninstall oniguruma");
+    assert!(!t.bin_dir().join("jq").exists());
+    assert_no_installed_symlinks(&t.root.path().join("prefix"));
+
+    assert_success(&t.zb(&["install", "jq"]), "zb install jq (reinstall)");
+    assert_success(
+        &t.run_binary("jq", &["--version"]),
+        "jq --version after reinstall",
+    );
+}
+
+#[test]
+#[ignore = "integration test"]
+fn test_list_installed_formulas() {
+    let t = TestEnv::new();
+
+    let output = t.zb(&["list"]);
+    assert_success(&output, "zb list (empty)");
+    assert_stdout_contains(&output, "No formulas installed");
+
+    assert_success(&t.zb(&["install", "jq"]), "zb install jq");
+
+    let output = t.zb(&["list"]);
+    assert_success(&output, "zb list");
+    assert_stdout_contains(&output, "jq");
+
+    assert_success(&t.zb(&["uninstall", "jq"]), "zb uninstall jq");
+    assert_success(&t.zb(&["uninstall", "oniguruma"]), "zb uninstall oniguruma");
+
+    let output = t.zb(&["list"]);
+    assert_success(&output, "zb list (empty)");
+    assert_stdout_contains(&output, "No formulas installed");
+}
+
+#[test]
+#[ignore = "integration test"]
+fn test_info_finds_installed_formula() {
+    let t = TestEnv::new();
+
+    let output = t.zb(&["info", "jq"]);
+    assert_success(&output, "zb info jq (not installed)");
+    assert_stdout_contains(&output, "not installed");
+
+    assert_success(&t.zb(&["install", "jq"]), "zb install jq");
+
+    let output = t.zb(&["info", "jq"]);
+    assert_success(&output, "zb info jq");
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(
+        stdout.contains("Name:") && !stdout.contains("not installed"),
+        "stdout: {stdout}"
+    );
+}
+
+#[test]
+#[ignore = "integration test"]
+fn test_gc_removes_unused_store_entries() {
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["gc"]), "zb gc (empty)");
+    assert_eq!(t.count_store_entries(), 0);
+
+    assert_success(&t.zb(&["install", "jq"]), "zb install jq");
+    let entries_before = t.count_store_entries();
+    assert!(entries_before > 0);
+
+    assert_success(&t.zb(&["uninstall", "jq"]), "zb uninstall jq");
+    assert_success(&t.zb(&["uninstall", "oniguruma"]), "zb uninstall oniguruma");
+    assert_eq!(t.count_store_entries(), entries_before);
+
+    assert_success(&t.zb(&["gc"]), "zb gc");
+    assert_eq!(t.count_store_entries(), 0);
+}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,5 +59,8 @@ jobs:
       - name: build (release)
         run: cargo build --workspace --release
       
-      - name: run tests
+      - name: run unit tests
         run: cargo test --workspace
+
+      - name: run integration tests
+        run: cargo test --package zb_cli --test integration -- --ignored
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,8 @@ out in the Discord to give us a heads up or open an issue first to discuss your 
 2. Make your changes and ensure, at the _least_:
    - Code is formatted: `cargo fmt --all`
    - No clippy warnings: `cargo clippy --workspace --all-targets -- -D warnings`
-   - Tests pass: `cargo test --workspace`
+   - Unit tests pass: `cargo test --workspace`
+   - Integration tests pass: `cargo test --workspace -- --ignored` (or `just test` for all tests)
 > [!NOTE] 
 > These will run in CI but it's best you clean up your code _before_ opening a PR to ensure a quick 
 > turnaround!
@@ -58,7 +59,7 @@ This project includes a `Justfile`, Install [just](https://github.com/casey/just
 - `just uninstall` Remove all zerobrew installations and configurations
 - `just fmt` Check code formatting
 - `just lint` Run clippy with strict warnings
-- `just test` Run all workspace tests
+- `just test` Run all workspace tests (unit & integration)
 
 Before creating a PR make sure you `build` your changes and `test` them.
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2564,6 +2564,7 @@ dependencies = [
  "tar",
  "tempfile",
  "tokio",
+ "walkdir",
  "wiremock",
  "zb_core",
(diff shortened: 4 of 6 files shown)


rbutoi:
> @rbutoi thanks!
> 
> > These tests are intentionally minimal
> 
> What's the intention? Ie why would we want minimal tests instead of validation of proper behavior

Apologies, I should've elaborated--I meant more to say that I tried to keep the duplication of validation to a minimum and not as coupled to internal details. For example, I didn't want this test to be strongly coupled to the exact output format of the `zb info` command (that would be better off in a unit test). Or, for the subsequent tests after the `reinstall` test, I didn't try and re-validate that running `jq` works.

The tests still do proper validation, IMO. Can you point to any missing coverage? (edit: e.g. https://github.com/lucasgelfond/zerobrew/issues/138 but I'll resolve that in another PR)

and thanks for the prompt reply!

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    fn bin_dir(&self) -> PathBuf {
+        self.root.path().join("prefix").join("bin")
+    }
+
+    fn count_store_entries(&self) -> usize {
+        std::fs::read_dir(self.root.path().join("store"))
+            .map(|r| r.count())
+            .unwrap()
+    }
```
cachebag, line 33: This function will panic if store dir doesn't exist yet. We should be handling `Result` or ensure dir exists first.

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,152 @@
...
+                "PATH",
+                format!(
+                    "{}:{}",
+                    self.bin_dir().display(),
+                    std::env::var("PATH").unwrap_or_default()
+                ),
+            )
+            .args(args)
+            .output()
+            .unwrap()
```
cachebag: You're using this way too liberally throughout the entire file. For integration tests we prefer that you use `expect()` with proper context and/or error handling

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,152 @@
...
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["install", "jq"]), "zb install jq");
+
+    let output = t.run_binary("jq", &["--version"]);
+    assert_success(&output, "jq --version");
+    assert!(String::from_utf8_lossy(&output.stdout).contains("jq-"));
+
+    assert_success(&t.zb(&["uninstall", "jq"]), "zb uninstall jq");
+    assert!(!t.bin_dir().join("jq").exists());
```
cachebag, line 141: This only checks that the binary doesn't exist, but it doesn't check if symlinks are cleaned up or that the formula was removed. Even for your "minimal" pass through, it should do that here.

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,152 @@
...
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["gc"]), "zb gc (empty)");
+    assert_success(&t.zb(&["install", "jq"]), "zb install jq");
+
+    let entries_before = t.count_store_entries();
+    assert!(entries_before > 0);
+
+    assert_success(&t.zb(&["uninstall", "jq"]), "zb uninstall jq");
+    assert_eq!(t.count_store_entries(), entries_before);
```
cachebag, line 208: This assertion is weak, you should check `== 0`. Or you could check if specific entries are gone.

cachebag (changes requested):
> The tests still do proper validation, IMO. Can you point to any missing coverage? 

They don't; the functionality of the binaries is missing (beyond ver checks), you don't test error paths, symlink correctness, etc... quite a few more things.

Which is _fine_ given if your goal is, like you said, just detecting regressions specifically with commands running without crashing. But with that being said, this PR does not resolve #119 and only hits a small piece of it.

rbutoi:
Addressed review comments in ~https://github.com/lucasgelfond/zerobrew/pull/139/changes/b3db6c0e6fdd30cb8e16500b4b6d985b93e55cd7 -- I'll squash this into the original `test(zb_cli)` commit after rounds of feedback. Please let me know if I should be squashing as I go.~ (edit: just read https://github.com/lucasgelfond/zerobrew/pull/129#pullrequestreview-3733404331, I've amended the commit) 

> > The tests still do proper validation, IMO. Can you point to any missing coverage?
> 
> They don't; the functionality of the binaries is missing (beyond ver checks), you don't test error paths, symlink correctness, etc... quite a few more things.
> 
> Which is _fine_ given if your goal is, like you said, just detecting regressions specifically with commands running without crashing. But with that being said, this PR does not resolve #119 and only hits a small piece of it.

fwiw, `ffmpeg` is tested exactly as the upstream Homebrew formula [does](https://github.com/Homebrew/homebrew-core/blob/baa32aa2d6b44bfbb2cbb6edd9407e39b8f327d0/Formula/f/ffmpeg.rb#L114). I've updated the `jq` test to be the same as the upstream as well.

edit: having read the tests in https://github.com/himudigonda/zerobrew/blob/31b8d836b6cb9281e2afd9b08786f517d54e801f/zb_io/src/install.rs#L722, I'm wondering roughly where the line should be drawn for what to validate where, and what's worth checking in both places

zb_cli/tests/integration.rs
rbutoi: Did the latter

zb_cli/tests/integration.rs
rbutoi: Sounds good. Added context with `expect()` except for utf-8 conversions of known test inputs.

zb_cli/tests/integration.rs
rbutoi: Added checks for symlink cleanup--what else should be checked for the formula being removed? e.g. peering into the database seems overly invasive / brittle

zb_cli/tests/integration.rs
rbutoi: I believe you mean for the assertion after (after gc)? But yes, changed to check `== 0`

rbutoi:
Fixed the [clippy error](https://github.com/lucasgelfond/zerobrew/actions/runs/21572251683/job/62153202864?pr=139), trying to understand the remaining [CI failures](https://github.com/lucasgelfond/zerobrew/actions/runs/21572251687/job/62153202848?pr=139). Looks like the fetched binaries fail to execute -- I'll try and do some testing on my own CI runners first before pushing more changes.

rbutoi:
Looking into the CI failures:

* `ffmpeg` had some runtime deps not available to the default github linux runner but could be fixed by `sudo apt-get update && sudo apt-get install -y libxcb-shape0 libxcb-shm0 libxcb-xfixes0`. This is overkill for a simple validation. I've added a simple `curl` test that works on all platforms.
* `jq` failed to execute because of `permission denied`. This was curious, and I tried `chmod +x`-ing it manually (from code with `fs::set_permissions`) before running (edit: as a simple check, not the solution) and that worked. Seems like Linux support was just merged in https://github.com/lucasgelfond/zerobrew/pull/51 and I might've found a bug in the patching logic at leaves binaries non-executable: https://github.com/lucasgelfond/zerobrew/pull/142 (I've rebased this PR on top of that as it's needed to pass).

rbutoi:
> This PR is steering out of scope. I don't mind the changes you made but do you think you could separate these?

This PR's scope hasn't changed, it's just rebased on top of another PR https://github.com/lucasgelfond/zerobrew/pull/142). The test revealed a real bug, fixed by bda21a1c55795c4efa0ee55c6b4d122130927ec4, that I've rebased this PR on top of the fix for (as I mentioned in my last comment). bda21a1c55795c4efa0ee55c6b4d122130927ec4 is not part of this PR, but is needed to pass the tests added here. I wish GitHub gave a way to make this more obvious - I'll edit the PR description.

The scope remains the CLI `integration.rs` test and the `init.rs` fixes to make the CLI testable.

cachebag:
> > This PR is steering out of scope. I don't mind the changes you made but do you think you could separate these?
> 
> 
> 
> This PR's scope hasn't changed, it's just rebased on top of another PR https://github.com/lucasgelfond/zerobrew/pull/142). The test revealed a real bug, fixed by bda21a1c55795c4efa0ee55c6b4d122130927ec4, that I've rebased this PR on top of the fix for (as I mentioned in my last comment). bda21a1c55795c4efa0ee55c6b4d122130927ec4 is not part of this PR, but is needed to pass the tests added here. I wish GitHub gave a way to make this more obvious - I'll edit the PR description.
> 
> 
> 
> The scope remains the CLI `integration.rs` test and the `init.rs` fixes to make the CLI testable.

Right but if we end up merging both PRs we're just gonna have dupe commits. Let's focus on one PR (ideally whatever needs the most attention first) and once it's merged to `main` we can just rebase this PR.

rbutoi:
You wouldn't merge the PRs (together) - the first is intended to go in independently of this (since it fixes an existing, latent bug discovered by this PR). After that, this one would be rebased onto `main` properly. I believe I've used this stack PR workflow before (with proper rebasing of course) and it didn't result in any duplicate commits.

cachebag:
> You wouldn't merge the PRs (together) - the first is intended to go in independently of this (since it fixes an existing, latent bug discovered by this PR). After that, this one would be rebased onto `main` properly. I believe I've used this stack PR workflow before (with proper rebasing of course) and it didn't result in any duplicate commits.

Maybe I wasn't clear initially, sorry- each PR has its own commit graph, if I merge #142, this PR will contain a commit (functionally, the _same_ commit that #142 has since they're the same hash) and that does not follow the workflow reviewers use.

The mechanics are right, yes, the commit will be dropped as a no-op since they're the same but GitHub doesn't know that until we merge this PR. It seems pedantic but we prefer things happen this way, as although this is not the case here- if someone else were to review this, they'd obviously want the PR to be contained to the scope outlined in the initial description. The note doesn't really do anything but add more friction. Each PR should be independently reviewable.

rbutoi:
Got it - I've rebased this back onto main with the understanding that one test will fail on Linux until https://github.com/lucasgelfond/zerobrew/pull/142 is merged (which is a real failure)

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    let output = t.zb(&["info", "jq"]);
+    assert_success(&output, "zb info jq");
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("Name:") && !stdout.contains("not installed"));
+}
+
+#[test]
+#[ignore = "integration test"]
+fn test_gc() {
```
cachebag: Nit: could we name this something better? Like `test_gc_removes_unused_store_entries`? And maybe follow a similar format for other tests where applicable

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    assert_success(&t.zb(&["install", "ffmpeg"]), "zb install ffmpeg");
+
+    let mp4out = t.root.path().join("video.mp4");
+    assert_success(
+        &t.run_binary(
+            "ffmpeg",
+            &[
+                "-filter_complex",
+                "testsrc=rate=1:duration=5",
```
cachebag, line 104: Magic numbers are bad news bears. Comments are fine in this case since we're running a binary, no need for constants

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,191 @@
...
+
+#[test]
+#[ignore = "integration test"]
+fn test_curl() {
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["install", "curl"]), "zb install curl");
+
+    let output = t.run_binary("curl", &["https://github.com"]);
+    assert_success(&output, "curl https://github.com");
```
cachebag: Why not use the `wiremock` server? It's already a dev dep. Otherwise, even using a local test server or a retry mechanism is better since network-dependent tests can be flaky

zb_cli/tests/integration.rs
cachebag: I think we should also check that the formula file itself is removed from the tap directory, and that the output of `brew list` doesn't include it anymore

zb_cli/tests/integration.rs
rbutoi: I believe I mentioned I got the test directly from the ffmpeg upstream formula? https://github.com/Homebrew/homebrew-core/blob/3076627c980d101ff02a720060c508433c44f293/Formula/f/ffmpeg.rb#L114

I'll leave that comment in the code then.

zb_cli/tests/integration.rs
rbutoi: > check that the formula file itself is removed from the tap directory

Do you mean, e.g. checking that `$prefix/opt/jq` is removed? Since that's covered

> I think we should also check that the formula file itself is removed from the tap directory, and that the output of `brew list` doesn't include it anymore

This is separately tested in the `list` test

zb_cli/tests/integration.rs
rbutoi: I'd like to respectfully push back here - I think setting up a wiremock (or even a local local server) is a bit overkill for this, given that these integration tests need network access to install `curl` in the first place. Sending a GET to a reliable and performant endpoint (which can by definition be reached easily from CI, since this *is* github) shouldn't be a stretch IMO.

I've changed to https://githubstatus.com for more perf/reliability

zb_cli/tests/integration.rs
rbutoi: Done

zb_cli/tests/integration.rs
cachebag: This is fair. Thank you!

zb_cli/tests/integration.rs
cachebag: I don't see the comment anywhere in your diff.

*Also, I'm concerned about a future reader of this code trying to understand why it's 5. Not everyone will know that this is how `ffmpeg` does this test upstream.

zb_cli/tests/integration.rs
cachebag: `test_list` only checks that installed formulas appear in the list - it doesn't test that uninstalled formulas are removed from the list. And by "formula file in tap directory" I mean the formula definition file (like homebrew's `tap/*.rb` files), not the installed files in `$prefix/opt`.

rbutoi:
> Few more notes. Also, I'm not sure if you just forgot to push the changes but I don't see any of the notes from my last review addressed? Just the one.

Just pushed https://github.com/lucasgelfond/zerobrew/pull/139/changes/50441edfbf8b24b335ac8aa1e732b226cb821ec3 to address the latest comments.

For the first round, I thought I had addressed them all -- the symlink checks, the GC entities # check, the `expect()`s, etc. Let me know if there's anything else still! This is the first integration test afaict (i.e. calling through external API which is the `zb` cli rather than tests linking into the code) so there's bound to be lots of changes.

zb_cli/tests/integration.rs
cachebag: Again, I don't see the changes in your diff. Are you ensuring the files are updated..?

zb_cli/tests/integration.rs
cachebag: Same here, I do not see the changes.

rbutoi: I check `== 0` on line 203 after running gc. Before gc, there should be the same number of store entries. Latest commit should be 50441ed (edit: before the push that this comment accompanied)

zb_cli/tests/integration.rs
rbutoi: > `test_list` only checks that installed formulas appear in the list - it doesn't test that uninstalled formulas are removed from the list.

Done in next push.

> And by "formula file in tap directory" I mean the formula definition file (like homebrew's `tap/*.rb` files), not the installed files in `$prefix/opt`.

I'm not sure where to find that--here's the tree structure of files created when I install a simple (the simplest I could find) package like GNU hello. Would it be any of these?

```
❱ cargo run --bin zb -- --root /tmp/h install hello
❱ eza -a --tree /tmp/h
.
├── cache
│   ├── blobs
│   │   └── d5eafb7348c3b7253714c061785a67f7878f36e953b454f04a80ec9c62482720.tar.gz
│   └── tmp
├── db
│   └── zb.sqlite3
├── locks
│   └── d5eafb7348c3b7253714c061785a67f7878f36e953b454f04a80ec9c62482720.lock
├── prefix
│   ├── bin
│   │   └── hello -> /tmp/h/prefix/Cellar/hello/2.12.2/bin/hello
│   ├── Cellar
│   │   └── hello
│   │       └── 2.12.2
│   │           ├── .brew
│   │           │   └── hello.rb
│   │           ├── AUTHORS
│   │           ├── bin
│   │           │   └── hello
│   │           ├── ChangeLog
│   │           ├── COPYING
│   │           ├── NEWS
│   │           ├── README
│   │           ├── sbom.spdx.json
│   │           ├── share
│   │           │   ├── info
│   │           │   │   └── hello.info
│   │           │   └── man
│   │           │       └── man1
│   │           │           └── hello.1
│   │           └── TODO
│   ├── include
│   ├── lib
│   ├── opt
│   │   └── hello -> /tmp/h/prefix/Cellar/hello/2.12.2
│   └── share
│       ├── info
│       │   └── hello.info -> /tmp/h/prefix/Cellar/hello/2.12.2/share/info/hello.info
│       └── man
│           └── man1
│               └── hello.1 -> /tmp/h/prefix/Cellar/hello/2.12.2/share/man/man1/hello.1
└── store
    └── d5eafb7348c3b7253714c061785a67f7878f36e953b454f04a80ec9c62482720
        └── hello
            └── 2.12.2
                ├── .brew
                │   └── hello.rb
                ├── AUTHORS
                ├── bin
                │   └── hello
                ├── ChangeLog
                ├── COPYING
                ├── NEWS
                ├── README
                ├── sbom.spdx.json
                ├── share
                │   ├── info
                │   │   └── hello.info
                │   └── man
                │       └── man1
                │           └── hello.1
                └── TODO
```

The only `.rb` files I see are in `store`, which is not the concern of this test (it is for the gc test)

zb_cli/tests/integration.rs
rbutoi: The comment is on line 96

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,211 @@
...
+
+#[test]
+#[ignore = "integration test"]
+#[cfg(target_os = "macos")] // GitHub Actions linux runner needs additional X11/XCB deps
+fn test_ffmpeg_formula() {
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["install", "ffmpeg"]), "zb install ffmpeg");
+
+    // From the upstream formula test: https://github.com/Homebrew/homebrew-core/blob/3076627c980d101ff02a720060c508433c44f293/Formula/f/ffmpeg.rb#L114
```
cachebag: Break the line here at `:`

zb_cli/tests/integration.rs
cachebag: @rbutoi Should be fine, don't worry about it.

zb_cli/tests/integration.rs
```diff
@@ -0,0 +1,211 @@
...
+#[ignore = "integration test"]
+fn test_gc_removes_unused_store_entries() {
+    let t = TestEnv::new();
+
+    assert_success(&t.zb(&["gc"]), "zb gc (empty)");
+    assert_eq!(t.count_store_entries(), 0);
+
+    assert_success(&t.zb(&["install", "jq"]), "zb install jq");
+    let entries_before = t.count_store_entries();
+    assert!(entries_before > 0);
```
cachebag, line 204: > I check `== 0` on line 203 after running gc. Before gc, there should be the same number of store entries. Latest commit should be [50441ed](https://github.com/lucasgelfond/zerobrew/commit/50441edfbf8b24b335ac8aa1e732b226cb821ec3) (edit: before the push that this comment accompanied)

No you don't?

rbutoi: As of 50441ed, it was line 203. On bfeebf3d7d53637afa6223e0e5e983dd41e83c07, it's line 210. Hope that clears it up.

rbutoi:
Pushed again, [a790baf](https://github.com/lucasgelfond/zerobrew/commit/a790bafd7d694defb4813b64128d4900817746ec). Please note that this will shift line numbers by 1

cachebag (approved):
Thanks! Going to re-open the issue since we aren't quite done yet but this is a great start.

PR #140, opened 2026-02-01
refactor(zb_cli): in `info`, make install timestamps human-readable

rbutoi:
The installation date in `zb info` is currently printed using the `std::fmt::Debug` trait as `SystemTime { tv_sec: xxx, tv_nsec: 0 }`.

### DIFF
--- a/zb_cli/src/commands/info.rs
+++ b/zb_cli/src/commands/info.rs
@@ -1,4 +1,5 @@
 use console::style;
+use std::time::{Duration, UNIX_EPOCH};
 
 pub fn execute(
     installer: &mut zb_io::install::Installer,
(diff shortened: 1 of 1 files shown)


rbutoi:
> We're adding an extra dep instead of just fixing the debug output? Nothing introduced from `chrono` can't be introduced using `std::time`?

Thanks, I did more digging and wasn't aware of `OffsetDateTime`.

cachebag (changes requested):
We're still pulling in a new dep that's simply not needed (at least not now). We can just do something like:

```rust
use std::time::{Duration, UNIX_EPOCH};

fn format_timestamp(timestamp: i64) -> String {
    let secs = timestamp.max(0) as u64;
    let dt = UNIX_EPOCH + Duration::from_secs(secs);

    match dt.duration_since(UNIX_EPOCH) {
        Ok(dur) => format!("{}s since epoch", dur.as_secs()),
        Err(_) => "invalid timestamp".to_string(),
    }
}
```

rbutoi:
SGTM

cachebag (changes requested):
Please update the respective call-sites.

rbutoi:
> Please update the respective call-sites.

sorry! pushed too soon, should've checked `check`

PR #141, opened 2026-02-01
chore(changelog): introduce changelog

cachebag:
Introduced our changelog.

Initial release shouldn't have any in-depth notes about features, we can just link them to the README.

~~We should not merge this until we develop a roadmap, @maria-rcks @lucasgelfond we can talk about that internally.~~
Roadmap will come down the line


maria-rcks (approved):
LGTM

PR #145, opened 2026-02-02
feat(zb_cli): support uninstalling multiple packages at once

rbutoi:
Also don't uninstall all packages when none are specified. ~This seems error-prone, and can come back as an `--all` flag if needed.~ edit: `--all` flag added

Tests can be added after https://github.com/lucasgelfond/zerobrew/pull/139

### DIFF
--- a/zb_cli/src/cli.rs
+++ b/zb_cli/src/cli.rs
@@ -34,7 +34,10 @@ pub enum Commands {
         no_link: bool,
     },
     Uninstall {
-        formula: Option<String>,
+        #[arg(required_unless_present = "all", num_args = 1..)]
+        formulas: Vec<String>,
+        #[arg(long)]
+        all: bool,
     },
     Migrate {
         #[arg(long, short = 'y')]
--- a/zb_cli/src/commands/uninstall.rs
+++ b/zb_cli/src/commands/uninstall.rs
@@ -2,43 +2,55 @@ use console::style;
 
 pub fn execute(
     installer: &mut zb_io::install::Installer,
-    formula: Option<String>,
+    formulas: Vec<String>,
+    all: bool,
 ) -> Result<(), zb_core::Error> {
-    match formula {
-        Some(name) => {
-            println!(
-                "{} Uninstalling {}...",
-                style("==>").cyan().bold(),
-                style(&name).bold()
-            );
-            installer.uninstall(&name)?;
-            println!(
-                "{} Uninstalled {}",
-                style("==>").cyan().bold(),
-                style(&name).green()
-            );
+    let formulas = if all {
+        let installed = installer.list_installed()?;
+        if installed.is_empty() {
+            println!("No formulas installed.");
+            return Ok(());
         }
-        None => {
-            let installed = installer.list_installed()?;
-            if installed.is_empty() {
-                println!("No formulas installed.");
-                return Ok(());
+        installed.into_iter().map(|k| k.name).collect()
+    } else {
+        formulas
+    };
+
+    println!(
+        "{} Uninstalling {}...",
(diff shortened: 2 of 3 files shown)


zb_cli/src/commands/uninstall.rs
```diff
@@ -2,43 +2,28 @@ use console::style;
...
 
-            println!("{} Uninstalled all packages", style("==>").cyan().bold());
+    for name in &formulas {
+        if formulas.len() > 1 {
+            print!("    {} {}...", style("○").dim(), name);
+        }
+        installer.uninstall(name)?;
+        if formulas.len() > 1 {
+            println!(" {}", style("✓").green());
         }
```
cachebag: So now we do..
```log
==> Uninstalling foo
==> Uninstalled foo
```

Is this intentional? UX isn't as clean here.

zb_cli/src/commands/uninstall.rs
cachebag, line 1: If the user runs `zb uninstall foo bar baz` and `bar` fails, the operation stops immediately and `baz` is never attempted. The current behavior is consistent with the old "uninstall all" flow, but could we collect errors and continue with the rest of the packages?

It would be good to see which packages failed (and why), and succeeded at the end.

zb_cli/src/cli.rs
```diff
@@ -34,7 +34,8 @@ pub enum Commands {
         no_link: bool,
     },
     Uninstall {
-        formula: Option<String>,
+        #[arg(required = true, num_args = 1..)]
```
cachebag: This is a breaking change for users..I haven't looked to deep but is there no way to avoid this?

Edit: i.e. If users relied on `zb uninstall` to clean up, they'll need to use a different approach (maybe pipe zb list into `xargs`?).

cachebag (changes requested):
Will be sure to look at #139 first but some quick comments for this one...

zb_cli/src/cli.rs
rbutoi: If not for this, we'd have to manually check if no packages were passed.

But yes, this *is* a breaking change and intentionally so. Previously, uninstall was a 0-or-1 package operation, now it's it's a 1+ package op.

> Edit: i.e. If users relied on zb uninstall to clean up, they'll need to use a different approach (maybe pipe zb list into xargs?).

Yeah - but I think we're early enough in the dev lifecycle to make these kind of judgements. A package manager that unconditionally destroys everything if you accidentally a word is a massive footgun :)

zb_cli/src/commands/uninstall.rs
rbutoi: I agree it's kind of redundant - but it's how things were already. I'll fix it.

cachebag: Thanks. I'm still getting up to speed on all of the code, so apologies if I'm asking too much. I still don't have a full grasp on who touched what

zb_cli/src/commands/uninstall.rs
rbutoi: Good call, done. That does leave a question about the API of this function though, since other commands return errors up to the caller and we're handling / reporting them here. Left a TODO there, let me know what you think.

rbutoi:
Pushed changes to remove redundant printing and allow uninstall failures to happen and continue

zb_cli/src/cli.rs
cachebag: @rbutoi An `--all` flag would be good, it's just as backwards to remove the logic completely.

cachebag (changes requested):
LGTM otherwise!

rbutoi:
Added `--all` flag

PR #107, opened 2026-01-29
refactor: move benchmark over to Justfile

maria-rcks:
### summary:
- migrated benchmark functionality from  `benchmark.sh` to `just bench`
- CSV format output
- quality of life: `--dry-run`, `--no-color`, `--output`, `--log` and pre-run validation 
- **included benchmark results on `results/` (ran on linux with`just bench --full results/`)**

### others:
- if no --format is set, selected from -o file.`ext`
- multi format export via `--full [dir]`
- defaults to `quick_packages` list ; use `--full` for all 100
- uses zb reset -y instead of rm -rf'ing /opt/zerobrew
- shows both cold/warm speedup per row

old:
```sh
$ ./benchmark.sh --help 
Usage: ./benchmark.sh [OPTIONS]

Options:
  -c, --count N     Number of packages to test (default: 100)
  --format FORMAT   Output format: text, json, or html (default: text)
  -o, --output FILE Write results to file instead of stdout
  --quick           Use quick package list (fewer deps)
  -h, --help        Show this help
```

new:
```sh
$ just bench --help
Usage: just bench [options]

Options:
  --quick              Test all quick packages (default, 22 packages)
  --full [DIR]         Test all 100 top Homebrew packages
                       Optionally specify DIR to output all formats to directory
  -c, --count N        Test first N packages from selected list
                       (quick packages by default, or from --full list)
  --format FORMAT      Output format: text (default), json, csv, or html
  -o, --output FILE    Write output to file instead of stdout (format inferred from extension)
  --no-color           Disable colored output
  --log FILE           Write install command logs to file
  --dry-run            Show what would be tested without running benchmarks
  --help               Show this help message
  ```
  
> AI-assisted: Planning and initial implementation with Claude Opus 4.5 / Sonnet 4.5, manually reviewed and tested.

PR #147, opened 2026-02-02
build: fix Justfile failing to run if `doas` is not installed

rbutoi:


cachebag:
Already being worked on in #107

PR #151, opened 2026-02-02
refactor(install.sh): better install / onboarding

maria-rcks:
**changes:**
- refactors install.sh with proper error handling, colors, spinner animation, and better help output
- adds --no-modify-path and -b/--binary flags to install.sh
- moves shell config setup from install.sh to 'zb init' command
- adds --no-modify-path flag to 'zb init' to skip shell config modification
> AI-assisted: Planning and initial implementation with Kimi K2.5, manually reviewed and tested.

image

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,8 +1,11 @@
-#!/bin/bash
-set -e
+#!/usr/bin/env bash
+set -euo pipefail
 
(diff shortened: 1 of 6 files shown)


maria-rcks:
@cachebag should be good now

PR #155, opened 2026-02-03
fix(zb_core): fix bottle selection to detect CPU arch on macOS

### LINKED ISSUE #78: bad CPU type in executable: mysql_secure_installation
Image

cachebag:
Previously, bottle selection always chose ARM64 bottles on macOS regardless of the actual CPU architecture. This caused "Bad CPU type in executable" errors when running binaries on Intel Macs.

Now we use compile-time `target_arch` detection to select the correct bottles.

Fixes #78


cachebag:
> LGTM, although ideally add tests for the initial selection maybe? just simple unit tests that we select the right thing based on arch? possibly overkill - I haven't tested this but the code LGTM

Since the logic uses compile-time `cfg!` macros, we can't test both architectures in a single test run. CI would need to run the test suite on both Intel and ARM64 macOS runners if we wanna test regressions and I believe Intel runner require a billing method on file. 

The tests I have will fail, however, on the respective architecture if the selection logic breaks so we have that at least

Issue #162, opened 2026-02-03
info & list more info


frederik-uni:
- Info should contain install information(where & how big) 
- info should contain dependencies
- info can contain description

- Would be great if list would contain a structured output & not just every installed package. for example a tree structure. I dont want to delete everything when cleaning up. maybe store it as a group. each package gets its on group. would allow to categorize packages and see who uses what easily. Would also allow some kind of session stuff like I want to find a fitting tool install 20 different packages & can just delete everything after without deleting everything

```
dav1d 1.5.3
ffmpeg 8.0.1_2
lame 3.100
libvpx 1.15.2
opus 1.6.1
sdl2 2.32.10
svt-av1 4.0.1
x264 r3222
x265 4.1
```
```
Name:       ffmpeg
Version:    8.0.1_2
Store key:  fb2b2a50ba2d
Installed:  15:08 (1 minutes ago)
```
```
❯ brew info ffmpeg
==> ffmpeg ✔: stable 8.0.1 (bottled), HEAD
Play, record, convert, and stream select audio and video codecs
https://ffmpeg.org/
Installed
/opt/homebrew/Cellar/ffmpeg/8.0.1_2 (284 files, 53.2MB) *
  Poured from bottle using the formulae.brew.sh API on 2026-02-03 at 15:09:05
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/f/ffmpeg.rb
License: GPL-2.0-or-later
==> Dependencies
Build: pkgconf ✔
Required: dav1d ✔, lame ✔, libvpx ✔, opus ✔, sdl2 ✔, svt-av1 ✔, x264 ✔, x265 ✔
```

PR #163, opened 2026-02-03
dep awareness

frederik-uni:
- keeps track of what was installed by what(explicitly/dep)
- deletes deps
- logs deps

ai was used


frederik-uni:
#162

frederik-uni:
I would love to know how I messed up the commit message ;D

frederik-uni:
the commit message got most likely messed up by amend and not ai. How do you want to test migration, which is the most likely point of failure. Tested it manually, but the rest is simple printing

PR #165, opened 2026-02-04
chore: update dependencies

maria-rcks:
`cargo update`


cachebag (approved):
LGTM. We should look into dependabot

PR #166, opened 2026-02-04
fix(zb_io): fix macOS git HTTPS remotes by patching hardcoded Homebrew paths in Mach-O binaries

### LINKED ISSUE #157: Git remote-https broken on macOS due to hardcoded paths
After installing git via zerobrew on macOS (Apple Silicon), HTTPS remote operations fail:

```
git: 'remote-https' is not a git command. See 'git --help'.
fatal: remote helper 'https' aborted session
```

Running `git --exec-path` returns the correct zerobrew path, but the binary itself has `/opt/homebrew/opt/git/libexec/git-core` baked in from the original Homebrew bottle.

## Environment

- macOS 15 (Sequoia), Apple Silicon
- zerobrew prefix: `/opt/zerobrew/prefix`
- git version: 2.52.0
- git installed via: `zb install git`

## What I checked

The helper binary exists and works fine when called directly:

```bash
$ ls -la /opt/zerobrew/prefix/opt/git/libexec/git-core/git-remote-https
lrwxr-xr-x  1 mike  wheel  15 Nov 17 11:38 git-remote-https -> git-remote-http

$ file /opt/zerobrew/prefix/opt/git/libexec/git-core/git-remote-http
Mach-O 64-bit executable arm64
```

The issue is git can't find it because it's looking in the wrong place.

## Workaround

Setting `GIT_EXEC_PATH` fixes it:

```bash
export GIT_EXEC_PATH="/opt/zerobrew/prefix/opt/git/libexec/git-core"
```

But this needs to be set everywhere - shell configs, launchd for GUI apps, etc. Not ideal.

## Likely cause

Same root cause as #154 (Linux). The macOS bottles also have hardcoded paths that aren't being relocated. The `@@HOMEBREW_PREFIX@@` placeholder handling might work, but absolute paths like `/opt/homebrew/...` baked into Mach-O binaries aren't getting patched.

I ran into this when Conductor (a macOS app that manages git worktrees) couldn't create workspaces - it doesn't inherit shell environment variables, so it was hitting the broken paths directly.

cachebag:
Closes #157 
~~WIP~~

Fixes a regression from #153 where HTTPS remote operations with git fail. Adds a test for this and patches Mach-O binaries.

### DIFF
--- a/zb_io/src/materialize.rs
+++ b/zb_io/src/materialize.rs
@@ -1013,6 +1013,35 @@ mod tests {
 
     #[test]
     #[cfg(target_os = "macos")]
+    fn test_patch_macho_skips_when_new_prefix_longer() {
+        let tmp = TempDir::new().unwrap();
+        let test_file = tmp.path().join("test_binary");
+
+        let old_prefix = "/opt/homebrew";
+        let new_prefix = "/opt/zerobrew/prefix";
+
+        let mut contents = Vec::new();
+        contents.extend_from_slice(b"\xfe\xed\xfa\xcf");
+        contents.extend_from_slice(b"some random data\0");
+        contents.extend_from_slice(old_prefix.as_bytes());
+        contents.extend_from_slice(b"/opt/git/libexec/git-core\0");
+        contents.extend_from_slice(b"more data\0");
+
+        let original = contents.clone();
+        fs::write(&test_file, &contents).unwrap();
+
+        let result = patch_macho_binary_strings(&test_file, new_prefix);
+        assert!(result.is_ok());
+
+        let patched = fs::read(&test_file).unwrap();
+        assert_eq!(
+            patched, original,
+            "binary should be unchanged when new prefix is longer than old"
+        );
+    }
+
+    #[test]
+    #[cfg(target_os = "macos")]
     fn test_patch_text_file_strings() {
         let tmp = TempDir::new().unwrap();
         let test_file = tmp.path().join("test_script.sh");


cachebag:
Works for me on macOS. @lucasgelfond or @mikecfisher, feel free to test if you'd like.

Made sure to also add a test to ensure that the binary is left unchanged (i.e. uncorrupted) even if the new prefix is longer than the old one.

zb_io/src/materialize.rs
```diff
@@ -139,8 +139,217 @@ fn find_bottle_content(store_entry: &Path, name: &str, version: &str) -> Result<
...
+    Ok(())
+}
+
+/// Patch hardcoded Homebrew paths in Mach-O binary data sections.
+/// This handles paths like /opt/homebrew/opt/git/libexec/git-core that are baked into binaries.
+#[cfg(target_os = "macos")]
+fn patch_macho_binary_strings(
+    path: &Path,
+    new_prefix: &str,
+    _new_cellar: &str,
```
maria-rcks: `_new_cellar` never used?

zb_io/src/materialize.rs
```diff
@@ -139,8 +139,217 @@ fn find_bottle_content(store_entry: &Path, name: &str, version: &str) -> Result<
...
+        Ok(c) => c,
+        Err(_) => return Ok(()),
+    };
+
+    let homebrew_prefixes = [
+        "/opt/homebrew",
+        "/usr/local/Homebrew",
+        "/usr/local",
+        "/home/linuxbrew/.linuxbrew",
+    ];
```
maria-rcks: we might want to make this a constant

zb_io/src/materialize.rs
```diff
@@ -139,8 +139,217 @@ fn find_bottle_content(store_entry: &Path, name: &str, version: &str) -> Result<
...
+            .map_err(|e| Error::StoreCorruption {
+                message: format!("failed to write temp file: {e}"),
+            })?;
+        drop(temp_file);
+
+        fs::rename(&temp_path, path).map_err(|e| Error::StoreCorruption {
+            message: format!("failed to rename temp file: {e}"),
+        })?;
+
+        let _ = std::process::Command::new("codesign")
```
maria-rcks: error handling?

zb_io/src/materialize.rs
```diff
@@ -139,8 +139,217 @@ fn find_bottle_content(store_entry: &Path, name: &str, version: &str) -> Result<
...
+        "/usr/local",
+        "/home/linuxbrew/.linuxbrew",
+    ];
+
+    let mut new_content = content.clone();
+    let mut changed = false;
+
+    new_content = new_content
+        .replace("@@HOMEBREW_PREFIX@@", new_prefix)
+        .replace("@@HOMEBREW_CELLAR@@", new_cellar);
```
maria-rcks: we should consider all the @@ variables homebrew uses [keg_relocate.rb](https://github.com/Homebrew/brew/blob/main/Library/Homebrew/keg_relocate.rb)
```
@@HOMEBREW_REPOSITORY@@
@@HOMEBREW_LIBRARY@@
@@HOMEBREW_PERL@@
@@HOMEBREW_JAVA@@
```
> this can be done in another pr tho.

maria-rcks (changes requested):
minor things

zb_io/src/materialize.rs
```diff
@@ -139,8 +139,217 @@ fn find_bottle_content(store_entry: &Path, name: &str, version: &str) -> Result<
...
+                if i + old_bytes.len() < contents.len() && contents[i + old_bytes.len()] != 0 {
+                    if i + old_bytes.len() < contents.len() && contents[i + old_bytes.len()] == b'/'
+                    {
+                    } else {
+                        null_terminated = false;
+                    }
+                }
+
+                if null_terminated
+                    || i + old_bytes.len() < contents.len() && contents[i + old_bytes.len()] == b'/'
```
maria-rcks: consider:

```patch
diff --git a/zb_io/src/materialize.rs b/zb_io/src/materialize.rs
index b0ea25e..fcc2920 100644
--- a/zb_io/src/materialize.rs
+++ b/zb_io/src/materialize.rs
@@ -290,18 +290,10 @@ fn patch_macho_binary_strings(
             }

             if contents[i..i + old_bytes.len()] == *old_bytes {
-                let mut null_terminated = true;
-                if i + old_bytes.len() < contents.len() && contents[i + old_bytes.len()] != 0 {
-                    if i + old_bytes.len() < contents.len() && contents[i + old_bytes.len()] == b'/'
-                    {
-                    } else {
-                        null_terminated = false;
-                    }
-                }
+                let next = contents.get(i + old_bytes.len()).copied();
+                let ends_cleanly = matches!(next, None | Some(0) | Some(b'/'));

-                if null_terminated
-                    || i + old_bytes.len() < contents.len() && contents[i + old_bytes.len()] == b'/'
-                {
+                if ends_cleanly {
                     contents[i..i + new_bytes.len()].copy_from_slice(new_bytes);

                     if new_bytes.len() < old_bytes.len() {
```

zb_io/src/materialize.rs
cachebag: Saved us the time and did it for Linux too. Thank you!

PR #168, opened 2026-02-04
fix: align download concurrency and chunked fallback

maria-rcks:
**changes:**
- validate range support before chunked downloads and fall back to non-chunked when needed
- try alternate mirrors for chunked downloads
- scope GHCR auth tokens per formula to avoid cross-package reuse
- align download concurrency defaults/tests and drop stale resolver comment
> AI-assisted: Planning and initial implementation with gpt-5.2-codex, manually reviewed and tested.

### DIFF
--- a/zb_core/src/resolve.rs
+++ b/zb_core/src/resolve.rs
@@ -70,7 +70,6 @@ fn compute_closure(
         for dep in deps {
             // Skip dependencies that aren't in the formulas map
             // (they were filtered out due to missing bottles for this platform)
-            // FIXME: we should probably just ignore these dependencies instead of skipping the entire formula
             if !formulas.contains_key(&dep) {
                 continue;
             }


zb_io/src/download.rs
```diff
@@ -286,19 +286,33 @@ impl Downloader {
...
-                client: &self.client,
-                token_cache: &self.token_cache,
-                url: primary_url,
-                expected_sha256,
-                name,
-                progress,
-                file_size: size,
-                global_semaphore: &semaphore,
-            };
+            let mut all_urls = Vec::new();
```
cachebag, line 289: Could we log which mirror succeeded/failed to help with debugging down the line? Could be helpful when investigating download issues.

(No changes needed, just wanna highlight this for the future when/if we implement logging)

zb_core/src/resolve.rs
```diff
@@ -70,7 +70,6 @@ fn compute_closure(
         for dep in deps {
             // Skip dependencies that aren't in the formulas map
             // (they were filtered out due to missing bottles for this platform)
-            // FIXME: we should probably just ignore these dependencies instead of skipping the entire formula
```
cachebag, line 73: Why was this removed?

cachebag (changes requested):
LGTM! Wonderful PR

zb_io/src/download.rs
maria-rcks: we need to implement logging, agree

zb_core/src/resolve.rs
maria-rcks:
```rs
if !formulas.contains_key(&dep) {
    continue;
}
```
already ignores the deps?

cachebag: Oh duh, sorry I misread the context around the file

PR #174, opened 2026-02-05
fix(zb_cli): restore homebrew installation suggestion on install errors

cachebag:
A complete blunder on my part- when reviewing #129 I completely missed that the `suggest_homebrew` functionality on errors was removed.

Nothing broke, but I think that it's helpful to users and completely my fault that I let that removal slip through.


maria-rcks (approved):
LGTM

PR #180, opened 2026-02-05
fix(install): handle empty init_args array on bash 3.2

shkm:
## Summary

Fixes "unbound variable" error when running install script on macOS with default bash 3.2.

## Problem

macOS ships with bash 3.2 (due to GPLv3 licensing), which has a quirk where expanding an empty array with `"${array[@]}"` triggers an "unbound variable" error even when the array is declared, when `set -u` is enabled.

```
bash: line 146: init_args[@]: unbound variable
```

This was fixed in bash 4.4+, but since the install script targets macOS users who likely have bash 3.2, we need to handle this case.

## Solution

Use the `${array[@]+"${array[@]}"}` pattern which safely expands to nothing when the array is empty/unset.

## Testing

- Tested on macOS with bash 3.2 - error no longer occurs
- The installation completes successfully including the `zb init` step

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -143,7 +143,7 @@ zb_init() {
         init_args+=("--no-modify-path")
     fi
 
-    "$zb_path" init "${init_args[@]}" >/dev/null 2>&1 || error_exit "Failed to initialize zerobrew"
+    "$zb_path" init ${init_args[@]+"${init_args[@]}"} >/dev/null 2>&1 || error_exit "Failed to initialize zerobrew"
 }
 
 print_logo() {


shkm:
> Please do not erase the PR template and replace it with AI generated slop.

Heya,

That's fair. I should note that I created the PR directly from Claude, as I thought the change was small enough not to warrant much extra, particularly given this project's LLM statement.

Since I haven't followed those instructions, and I don't intend to continue using this anyway, I'll go ahead and close.

cachebag:
> > Please do not erase the PR template and replace it with AI generated slop.
> 
> Heya,
> 
> That's fair. I should note that I created the PR directly from Claude, as I thought the change was small enough not to warrant much extra, particularly given this project's LLM statement.

Fair enough but this is one of those "if I make an exception for one, I should make an exception for all" kind of situations. It's quite pedantic but it's the guideline we must enforce.

Thanks anyways!

shkm:
> > > Please do not erase the PR template and replace it with AI generated slop.
> > 
> > 
> > Heya,
> > That's fair. I should note that I created the PR directly from Claude, as I thought the change was small enough not to warrant much extra, particularly given this project's LLM statement.
> 
> Fair enough but this is one of those "if I make an exception for one, I should make an exception for all" kind of situations. It's quite pedantic but it's the guideline we must enforce.
> 
> Thanks anyways!

I get it. No hard feelings by any means.

PR #181, opened 2026-02-05
fix(bottle): use bare OS names for Intel Mac bottle tags

sgrams:
This PR is to enable support for Intel-based Macs.

Homebrew uses bare OS version names (e.g. "sonoma") for Intel Mac bottles, not "x86_64_" prefixed names. Update tag matching and fallback logic accordingly.

This PR has been written with the use of Claude Opus 4.6.
Tested on macOS Sequoia 15.7.3 (Intel) and Arch Linux (x86_64).

#

- [x] I have disclosed in this PR if AI assistance was used to generate any code, documentation, or other contributions. I understand that if I do not adhere to this standard, my PR wil be closed, even without explanation.

### DIFF
--- a/zb_cli/src/commands/run.rs
+++ b/zb_cli/src/commands/run.rs
@@ -148,6 +148,8 @@ mod tests {
     fn get_test_bottle_tag() -> &'static str {
         if cfg!(target_os = "linux") {
             "x86_64_linux"
+        } else if cfg!(target_arch = "x86_64") {
+            "sonoma"
         } else {
             "arm64_sonoma"
         }


zb_io/src/extraction/patch/macos.rs
cachebag, line 1: This has nothing to do with Intel Mac support? Can you remove it.

zb_cli/tests/integration.rs
```diff
@@ -16,6 +16,7 @@ impl TestEnv {
         let zb = env!("CARGO_BIN_EXE_zb");
         Command::new(zb)
             .env("ZEROBREW_ROOT", self.root.path())
+            .env("ZEROBREW_PREFIX", self.root.path().join("prefix"))
```
cachebag, line 22: Could your provide context on why this is needed? Is it required for tests to pass on Intel? We haven't setup CI for Intel yet so I am not sure.

sgrams: Yes, integration tests were failing for me (run on pretty fresh install of macOS Sequoia)

zb_io/src/extraction/patch/macos.rs
sgrams: sure, let me drop it.

zb_cli/tests/integration.rs
cachebag: Okay can you please leave a comment on why we're including this new variable?

sgrams: I added a comment

PR #193, opened 2026-02-08
fix(zb_io, zb_cli, install): harden install/gc state and runtime defaults

maria-rcks:

#

changes:
- fixes store refcount leaks on reinstall/upgrade and adds DB regression tests
- persists successful installs even when a later package in the same run fails
- removes stale `store_refs` rows during GC
- infers runtime prefix for `zb run` from Cellar path when `ZEROBREW_PREFIX` is unset
- avoids panic paths in downloader TLS/native cert setup and timestamp writes
- rejects `--concurrency 0` in CLI
- fixes bash completions dir path expansion and install.sh unknown-option handling

> AI-assisted: planning + initial implementation with gpt-5.3-codex, manually reviewed and tested.

### DIFF
--- a/zb_io/src/installer/install.rs
+++ b/zb_io/src/installer/install.rs
@@ -278,9 +278,12 @@ impl Installer {
                 Ok(download) => {
                     let idx = download.index;
                     let (formula, bottle) = &to_install[idx];
+                    let processed_name = formula.name.clone();
+                    let processed_version = formula.effective_version();
+                    let processed_store_key = bottle.sha256.clone();
 
                     report(InstallProgress::UnpackStarted {
-                        name: formula.name.clone(),
+                        name: processed_name.clone(),
                     });
 
                     // Try extraction with retry logic for corrupted downloads
@@ -298,8 +301,8 @@ impl Installer {
                     // Materialize to cellar
                     // Use effective_version() which includes rebuild suffix if applicable
                     let keg_path = match self.cellar.materialize(
-                        &formula.name,
-                        &formula.effective_version(),
+                        &processed_name,
+                        &processed_version,
                         &store_entry,
                     ) {
                         Ok(path) => path,
@@ -310,22 +313,30 @@ impl Installer {
                     };
 
                     report(InstallProgress::UnpackCompleted {
-                        name: formula.name.clone(),
+                        name: processed_name.clone(),
                     });
 
                     // Link executables if requested
                     let linked_files = if link {
                         report(InstallProgress::LinkStarted {
-                            name: formula.name.clone(),
+                            name: processed_name.clone(),
                         });
                         match self.linker.link_keg(&keg_path) {
                             Ok(files) => {
                                 report(InstallProgress::LinkCompleted {
-                                    name: formula.name.clone(),
+                                    name: processed_name.clone(),
                                 });
                                 files
                             }
                             Err(e) => {
+                                Self::cleanup_failed_install(
+                                    &self.linker,
+                                    &self.cellar,
+                                    &processed_name,
+                                    &processed_version,
+                                    &keg_path,
+                                    true,
+                                );
                                 error = Some(e);
                                 continue;
                             }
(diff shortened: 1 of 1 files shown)


zb_io/src/storage/db.rs
```diff
@@ -166,29 +178,66 @@ impl<'a> InstallTransaction<'a> {
...
+                self.tx
+                    .execute(
+                        "INSERT INTO store_refs (store_key, refcount) VALUES (?1, 1)
+                         ON CONFLICT(store_key) DO UPDATE SET refcount = refcount + 1",
+                        params![store_key],
+                    )
+                    .map_err(|e| Error::StoreCorruption {
+                        message: format!("failed to increment store ref: {e}"),
+                    })?;
+            }
```
cachebag, line 233: This is identical to lines 230-238. If someone fixes a bug and/or adjusts the code in general in one arm and forgets the other, refcounting silently breaks. 

Please collapse this to something like:
```rust
match previous_store_key.as_deref() {
    Some(previous) if previous == store_key => {}
    other => {
        if let Some(previous) = other {
            // decrement old key
        }
        // increment new key (single copy)
    }
}
```

zb_io/src/network/download/mod.rs
```diff
@@ -127,28 +127,32 @@ struct CachedToken {
...
+        let _ = root_store.add(cert);
     }
 
-    rustls::ClientConfig::builder_with_provider(provider.into())
-        .with_safe_default_protocol_versions()
-        .expect("failed to set protocol versions")
-        .with_root_certificates(root_store)
-        .with_no_client_auth()
+    let builder = rustls::ClientConfig::builder_with_provider(provider.into());
+    let builder = builder.with_safe_default_protocol_versions().ok()?;
```
cachebag: This silently swallows the error and returns `None`. When this happens, `create_isolated_client` skips custom TLS entirely. It won't crash, but there's zero signal to the user that TLS setup failed. 

Add at least an `eprintln!` or before returning None so failures are diagnosable. (another reason we need to setup logging)

zb_io/src/network/download/mod.rs
```diff
@@ -127,28 +127,32 @@ struct CachedToken {
...
 
-fn build_rustls_config() -> rustls::ClientConfig {
+fn build_rustls_config() -> Option<rustls::ClientConfig> {
     let provider = rustls::crypto::aws_lc_rs::default_provider();
 
     let mut root_store = rustls::RootCertStore::empty();
 
-    for cert in rustls_native_certs::load_native_certs().expect("failed to load native certs") {
-        root_store.add(cert).ok();
+    let cert_result = rustls_native_certs::load_native_certs();
```
cachebag, line 135: This has an `.errors` field (since `rustls_native_certs` 0.8+) that's completely ignored. If all certs fail to load, you get an empty root store and every HTTPS request will fail with an opaque TLS error. Try to log `cert_result.errors` if non-empty.

zb_io/src/storage/db.rs
```diff
@@ -166,29 +178,66 @@ impl<'a> InstallTransaction<'a> {
...
+            .unwrap_or(0);
+
+        let previous_store_key: Option<String> = self
+            .tx
+            .query_row(
+                "SELECT store_key FROM installed_kegs WHERE name = ?1",
+                params![name],
+                |row| row.get(0),
+            )
+            .ok();
```
cachebag: `.ok()` collapses all errors to `None`, not just “no rows,” so real DB failures get treated like a fresh install and can leak a refcount. This matches the existing `record_uninstall` pattern, but since this is new code it’s worth noting that `rusqlite::optional()` exists to avoid that footgun.

zb_io/src/installer/install.rs
```diff
@@ -350,12 +339,52 @@ impl Installer {
...
-                        version: formula.effective_version(),
-                        store_key: bottle.sha256.clone(),
-                        linked_files,
-                    });
+                    let processed_name = formula.name.clone();
+                    let processed_version = formula.effective_version();
+                    let processed_store_key = bottle.sha256.clone();
+                    let processed_links = linked_files;
+
+                    // Persist successful package immediately so one later failure
```
cachebag, line 355: Hmm if `record_install` or `record_linked_file` errors, the cellar extraction and symlinks from lines 300-335 are already on disk. You get orphaned files with no DB record. 

If you don't want to fix it you don't have to (the old code had the same problem differently), but worth noting that "persist immediately" doesn't fully protect against this.

install.sh
```diff
@@ -182,7 +182,7 @@ while [[ $# -gt 0 ]]; do
         fi
         ;;
     *)
-        error_exit "Unknown option '%s'" "$1"
+        error_exit "Unknown option '$1'"
```
cachebag: Nit: `printf`-style %s was slightly safer against shell metacharacters in $1.

PR #196, opened 2026-02-08
fix(zb_cli): fish support & init shell block

maria-rcks:

#

> Initial plan and implementation by gpt-5.3-codex

- Adds support for fix shell
- Removes CA bundle override (dk why is needed) 
- Handles shell config in a block that can be updated (no longer checking if `# zerobrew` just exists)

### DIFF
--- a/zb_cli/src/init.rs
+++ b/zb_cli/src/init.rs
@@ -208,24 +208,6 @@ export ZEROBREW_ROOT={root}
 export ZEROBREW_PREFIX={prefix}
 export PKG_CONFIG_PATH="$ZEROBREW_PREFIX/lib/pkgconfig:${{PKG_CONFIG_PATH:-}}"
 
-# SSL/TLS certificates (only if ca-certificates is installed)
-if [ -f "$ZEROBREW_PREFIX/opt/ca-certificates/share/ca-certificates/cacert.pem" ]; then
-  export CURL_CA_BUNDLE="$ZEROBREW_PREFIX/opt/ca-certificates/share/ca-certificates/cacert.pem"
-  export SSL_CERT_FILE="$ZEROBREW_PREFIX/opt/ca-certificates/share/ca-certificates/cacert.pem"
-elif [ -f "$ZEROBREW_PREFIX/etc/ca-certificates/cacert.pem" ]; then
-  export CURL_CA_BUNDLE="$ZEROBREW_PREFIX/etc/ca-certificates/cacert.pem"
-  export SSL_CERT_FILE="$ZEROBREW_PREFIX/etc/ca-certificates/cacert.pem"
-elif [ -f "$ZEROBREW_PREFIX/share/ca-certificates/cacert.pem" ]; then
-  export CURL_CA_BUNDLE="$ZEROBREW_PREFIX/share/ca-certificates/cacert.pem"
-  export SSL_CERT_FILE="$ZEROBREW_PREFIX/share/ca-certificates/cacert.pem"
-fi
-
-if [ -d "$ZEROBREW_PREFIX/etc/ca-certificates" ]; then
-  export SSL_CERT_DIR="$ZEROBREW_PREFIX/etc/ca-certificates"
-elif [ -d "$ZEROBREW_PREFIX/share/ca-certificates" ]; then
-  export SSL_CERT_DIR="$ZEROBREW_PREFIX/share/ca-certificates"
-fi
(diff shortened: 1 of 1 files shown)


maria-rcks:
there!

zb_cli/src/init.rs
```diff
@@ -127,73 +127,87 @@ pub fn run_init(root: &Path, prefix: &Path, no_modify_path: bool) -> Result<(),
...
 }
 
+const ZB_BLOCK_START: &str = "# >>> zerobrew >>>";
+const ZB_BLOCK_END: &str = "# <<< zerobrew <<<";
+
+fn upsert_managed_block(existing: &str, managed_block: &str) -> String {
+    if let Some(start_idx) = existing.find(ZB_BLOCK_START)
+        && let Some(end_rel_idx) = existing[start_idx..].find(ZB_BLOCK_END)
+    {
+        let end_idx = start_idx + end_rel_idx + ZB_BLOCK_END.len();
```
cachebag, line 137: After computing `end_idx`, you need to consume the trailing newline. It points right after the `# <<< zerobrew <<<` but doesn't consume the trailing `\n` that follows it. Every invocation adds another blank line.

zb_cli/src/init.rs
```diff
@@ -127,73 +127,87 @@ pub fn run_init(root: &Path, prefix: &Path, no_modify_path: bool) -> Result<(),
...
+            (home_zshrc, ShellConfigKind::Posix)
         }
     } else if shell.contains("bash") {
-        let bash_profile = format!("{}/.bash_profile", home);
-        if std::path::Path::new(&bash_profile).exists() {
-            bash_profile
-        } else {
-            format!("{}/.bashrc", home)
-        }
+        (format!("{}/.bashrc", home), ShellConfigKind::Posix)
```
cachebag, line 187: This is going to break for many macOS users. Unconditionally targeting `.bashrc` is not desirable because every Terminal.app window is a login shell that sources `.bash_profile` but won't source `.bashrc` unless `.bash_profile` explicitly does so. The old code preferred `.bash_profile` when it existed for this reason

zb_cli/src/init.rs
cachebag, line 1: > Removes CA bundle override (dk why is needed)

Why would you remove something without knowing why it should or shouldn't be removed? 

IIRC, `CURL_CA_BUNDLE`, `SSL_CERT_FILE`, and `SSL_CERT_DIR` were set so that tools running under zerobrew could find TLS certificates installed by ca-certificates. This fixed some issues regarding git remote operations failing.

zb_cli/src/init.rs
```diff
@@ -206,17 +220,52 @@ _zb_path_append() {{
...
+                prefix = prefix.display()
+            ),
+            ShellConfigKind::Fish => format!(
+                r#"
+# zerobrew
+set -gx ZEROBREW_DIR "{zerobrew_dir}"
+set -gx ZEROBREW_BIN "{zerobrew_bin}"
+set -gx ZEROBREW_ROOT "{root}"
+set -gx ZEROBREW_PREFIX "{prefix}"
+set -gx PKG_CONFIG_PATH "$ZEROBREW_PREFIX/lib/pkgconfig:$PKG_CONFIG_PATH"
```
cachebag, line 235: Nit: If `PKG_CONFIG_PATH` is not previously set, this produces a trailing colon (`/path:`) in fish. The POSIX version uses `${PKG_CONFIG_PATH:-}` to handle this gracefully. The fish equivalent would be something like:
```bash
set -q PKG_CONFIG_PATH; or set -gx PKG_CONFIG_PATH ""
set -gx PKG_CONFIG_PATH "$ZEROBREW_PREFIX/lib/pkgconfig" $PKG_CONFIG_PATH
```

Honestly you could simply prepend using fish list semantics rather than colon-separated strings (fish PATH-like variables use list items, not colons)

cachebag (changes requested):

zb_cli/src/init.rs
maria-rcks: It just kept breaking my system, but before it didn't consider already existing cerificates

cachebag: That's valid- maybe let's add a new commit to the init file to fix this? Just set them only if they're not already set. The black or white approach isn't the best here

maria-rcks: doing that right now :saluting_face:

PR #212, opened 2026-02-11
feat(zb): local source building fallback and `zb_core` modular restructure

### LINKED ISSUE #27: emacs: fails to start due to hard-coded /opt/homebrew paths

### LINKED ISSUE #40: Emacs doesn't work on macOS Sequoia

### LINKED ISSUE #48: stow binary has unresolved @@HOMEBREW_PERL@@ placeholder

### LINKED ISSUE #149: ncurses cant find terminfo
I migrated basically my whole system to zb, which has been great except for this issue:

ncurses gets information about terminal capabilities through infocmp, which sources its data from terminfo entries, but it's unable to do this for `$ zb install ncurses`. This is because, as per `$ man terminfo`:

The environment variable TERMINFO is checked first, for a terminal database containing the terminal description.
 - Next, ncurses looks in $HOME/.terminfo for a compiled description.
   - This is an optional feature which may be omitted entirely from the library, or limited to prevent accidental use by privileged applications.
 - Next, if the environment variable TERMINFO_DIRS is set, ncurses interprets the contents of that variable as a list of colon-separated pathnames of terminal databases to be searched.
   - An empty pathname (i.e., if the variable begins or ends with a colon, or contains adjacent colons) is interpreted as the system location /opt/homebrew/Cellar/ncurses/6.6/share/terminfo.
 - Finally, ncurses searches these compiled-in locations:
   - a list of directories (@@HOMEBREW_CELLAR@@/ncurses/6.6/share/terminfo), and
   - the system terminfo directory, /opt/homebrew/Cellar/ncurses/6.6/share/terminfo

the tl;dr is that the homebrew build of ncurses expects the terminfo data to be under the `/opt/homebrew/` prefix, when zb installs it under `/opt/zerobrew/prefix`. This can be pretty easily with
```sh
export TERMINFO_DIRS="/opt/zerobrew/prefix/Cellar/ncurses/6.6/share/"
```
in the proper context or simply
```sh
sudo mkdir -p /opt/homebrew/Cellar/ncurses/6.6
sudo ln -s /opt/zerobrew/prefix/Cellar/ncurses/6.6/share /opt/homebrew/Cellar/ncurses/6.6/share
```

im not sure if it's possible to actually fix the underlying issue as long as zb is using a different prefix and sharing homebrew's builds, but this was the only real piece of friction i had with switching to zb, loving it so far

debugging note: running `$ infocmp -D` will show the paths ncurses is looking at that exist, which for me after doing the symlink gives
```
/Users/tj/.terminfo
/opt/homebrew/Cellar/ncurses/6.6/share/terminfo
```

### LINKED ISSUE #152: Installing git on linux failed - libiconv
Really loving zb so far. Just installed the latest available code for a few ubuntu machines. I can install most of my packages, but encountered 2 issues:

1. libiconv (git dependency)
```
❯ zb install libiconv
==> Installing libiconv...
error: unsupported bottle for formula 'libiconv'
```

2. Installing a non core formula
This one, I'm not sure, this is probably intended
```
❯ zb install jstkdng/programs/ueberzugpp
==> Installing jstkdng/programs/ueberzugpp...
error: tap formula 'jstkdng/programs/ueberzugpp' is not supported (only homebrew/core)
```

Cheers

### LINKED ISSUE #154: Git binary has hardcoded Homebrew paths on Linux
On Linux, the git binary installed via zerobrew contains hardcoded paths to `/home/linuxbrew/.linuxbrew/opt/git` instead of using the actual zerobrew installation path. This causes git to look for helpers and config files in the wrong location.

```bash
$ strings /opt/zerobrew/prefix/Cellar/git/2.52.0_1/bin/git | grep linuxbrew
/home/linuxbrew/.linuxbrew/etc/gitattributes
/home/linuxbrew/.linuxbrew/etc/gitconfig
/home/linuxbrew/.linuxbrew/opt/git

$ git --exec-path
/home/linuxbrew/.linuxbrew/opt/git/libexec/git-core
```

Git should use paths relative to the zerobrew installation, e.g., `/opt/zerobrew/prefix/Cellar/git/2.52.0_1/libexec/git-core`

I think the Linux patching code in `zb_io/src/linux_patch.rs` only replaces `@@HOMEBREW_PREFIX@@` and `@@HOMEBREW_CELLAR@@` placeholders, but doesn't handle hardcoded absolute paths that Homebrew Linux bottles contain.

We could extend `patch_elf_placeholders()` to do byte-level string replacement of common Homebrew paths like `/home/linuxbrew/.linuxbrew` with the actual zerobrew prefix path.

### LINKED ISSUE #159: [Linux] - neovim - liblpeg.so: cannot open shared object file:
neovim install succeeds, with the correct libraries, but some symlink must be missing

```❯ zb install neovim
==> Installing neovim...
==> Resolving dependencies (10 packages)...
    libunistring 1.4.1
    gettext 1.0
    libuv 1.51.0
    lpeg 1.1.0
    luajit 2.1.1767980792                                                                                                                                                                                                                           luv 1.51.0-2                                                                                                                                                                                                                                    tree-sitter@0.25 0.25.10
    unibilium 2.1.2                                                                                                                                                                                                                                 utf8proc 2.11.3                                                                                                                                                                                                                                 neovim 0.11.6
==> Downloading and installing...                                                                                                                                                                                                                                                                                                                                                                                                                                                               ==> Installed 10 packages in 1.05s


 which nvim
/home/ubuntu/local/share/zerobrew/prefix/bin/nvim


nvim
nvim: error while loading shared libraries: /home/linuxbrew/.linuxbrew/opt/lpeg/lib/liblpeg.so: cannot open shared object file: No such file or directory


```

### LINKED ISSUE #209: Zerobrew reuses packages built for a different prefix
Hello all. This issue is the root cause for the following issues:

- #159 
- #154 
- #152
- #149
- #69
- #40
- #27

In summary, homebrew bottles are built for a specific `$HOMEBREW_PREFIX`. If the prefix is changed, the package ***must*** be rebuilt from source to reflect the new `$HOMEBREW_PREFIX`. Since the default `$ZEROBREW_PREFIX` does not match the default `$HOMEBREW_PREFIX`, zerobrew ***must*** rebuild the packages from source in order for most packages that use `ldd` as a interpreter to work as expected since paths are baked in at link-time. 

See [the Homebrew project's guide on using a different prefix](https://docs.brew.sh/Installation#untar-anywhere-unsupported) for notes on implementation.

Since needing to build each bottle at install-time will indubitably slow down zerobrew, here are what I see the solutions to be:

1. Change the default `$ZEROBREW_PREFIX` to match `$HOMEBREW_PREFIX`
    * Pro: No need to rebuild packages
    * Con: Zerobrew cannot co-exist with an existing brew install, making a gradual migration impossible.
2. Build all packages on a zerobrew-dedicated CDN
    * Pro: Stays fast, and compatible.
    * Con: Need to maintain independent infrastructure
3. Accept that users will need to rebuild packages
    * Pro: Will never break, compatible with all upstream formulae.
    * Con: Slow!

I believe that solution #2 is the best.

cachebag:

#

~~WIP draft~~ PR for work discussed in #209 

This PR adds a local source build fallback so zerobrew can compile packages from source using Homebrew's Ruby formula files when no pre-built bottle is available for the platform. Additionally, refactors `zb_core` to follow arch conventions applied to `zb_cli` and `zb_io` 

Fixes #209
Fixes #159
Fixes #154
Fixes #152
Fixes #149
Fixes #48
Fixes #40
Fixes #27

### DIFF
--- a/zb_io/src/build/source.rs
+++ b/zb_io/src/build/source.rs
@@ -1,9 +1,9 @@
 use std::path::{Path, PathBuf};
 
-use sha2::{Digest, Sha256};
 use tokio::fs;
 use zb_core::Error;
 
+use crate::checksum::verify_sha256_bytes;
 use crate::extraction::extract_tarball;
 
 pub async fn download_and_extract_source(
@@ -14,9 +14,7 @@ pub async fn download_and_extract_source(
     let tarball_path = work_dir.join("source.tar.gz");
     download_source(url, &tarball_path).await?;
 
-    if let Some(expected) = expected_checksum {
-        verify_checksum(&tarball_path, expected).await?;
-    }
+    verify_checksum(&tarball_path, expected_checksum, url).await?;
 
     let src_dir = work_dir.join("src");
     fs::create_dir_all(&src_dir)
@@ -62,23 +60,18 @@ async fn download_source(url: &str, dest: &Path) -> Result<(), Error> {
     })
 }
 
-async fn verify_checksum(path: &Path, expected: &str) -> Result<(), Error> {
+async fn verify_checksum(path: &Path, expected: Option<&str>, url: &str) -> Result<(), Error> {
     let bytes = fs::read(path).await.map_err(|e| Error::FileError {
         message: format!("failed to read tarball for checksum: {e}"),
     })?;
 
-    let mut hasher = Sha256::new();
-    hasher.update(&bytes);
-    let actual = format!("{:x}", hasher.finalize());
-
-    if actual != expected.to_lowercase() {
-        return Err(Error::ChecksumMismatch {
-            expected: expected.to_string(),
-            actual,
-        });
-    }
-
-    Ok(())
+    verify_sha256_bytes(&bytes, expected).map_err(|e| match e {
+        Error::ChecksumMismatch { .. } => e,
+        Error::InvalidArgument { message } => Error::InvalidArgument {
+            message: format!("invalid source checksum for '{url}': {message}"),
+        },
+        other => other,
+    })
 }
 
 async fn find_source_root(src_dir: &Path) -> Result<PathBuf, Error> {
--- a/zb_io/src/checksum.rs
+++ b/zb_io/src/checksum.rs
@@ -1,0 +1,88 @@
+use sha2::{Digest, Sha256};
+use zb_core::Error;
+
+/// Verify the SHA-256 checksum of a byte slice.
+///
+/// When `expected_sha256` is `None` the check is skipped (caller opted out).
+/// When `Some`, the value must be a 64-character hex string; mismatches and
+/// malformed expectations are returned as typed errors.
+pub fn verify_sha256_bytes(bytes: &[u8], expected_sha256: Option<&str>) -> Result<(), Error> {
+    let Some(expected_sha256) = expected_sha256 else {
+        return Ok(());
+    };
+
+    let expected = normalize_sha256(expected_sha256)?;
+
+    let mut hasher = Sha256::new();
+    hasher.update(bytes);
+    let actual = format!("{:x}", hasher.finalize());
+
+    if actual != expected {
+        return Err(Error::ChecksumMismatch { expected, actual });
+    }
+
+    Ok(())
+}
+
+fn normalize_sha256(input: &str) -> Result<String, Error> {
+    let normalized = input.trim().to_lowercase();
+
+    if normalized.len() != 64 {
+        return Err(Error::InvalidArgument {
+            message: format!(
+                "invalid sha256 checksum: expected 64 hex chars, got {}",
+                normalized.len()
+            ),
+        });
+    }
+
+    if !normalized.bytes().all(|b| b.is_ascii_hexdigit()) {
+        return Err(Error::InvalidArgument {
+            message: "invalid sha256 checksum: must contain only hex characters".to_string(),
+        });
+    }
+
+    Ok(normalized)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn skips_verification_when_none() {
+        assert!(verify_sha256_bytes(b"anything", None).is_ok());
+    }
+
+    #[test]
+    fn accepts_valid_checksum() {
+        // SHA-256 of b"hello"
+        let expected = "2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824";
+        assert!(verify_sha256_bytes(b"hello", Some(expected)).is_ok());
+    }
+
+    #[test]
+    fn accepts_uppercase_and_whitespace() {
+        let expected = " 2CF24DBA5FB0A30E26E83B2AC5B9E29E1B161E5C1FA7425E73043362938B9824 ";
+        assert!(verify_sha256_bytes(b"hello", Some(expected)).is_ok());
+    }
+
+    #[test]
+    fn rejects_invalid_length() {
+        let err = verify_sha256_bytes(b"hello", Some("abc")).unwrap_err();
+        assert!(matches!(err, Error::InvalidArgument { .. }));
+    }
+
+    #[test]
+    fn rejects_non_hex() {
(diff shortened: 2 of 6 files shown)


cachebag:
Okay this is ready. Biggest thing worth noting here is the Ruby shim approach. Rather than reimplementing Homebrew's build DSL in Rust (which would be a nightmare to maintain as Homebrew evolves), we execute the actual `.rb` formula files through a controlled Ruby environment that mimics just enough of Homebrew's class hierarchy to make `install` methods run correctly. The shim defines a `Formula` base class with zerobrew-compatible paths, so when a formula calls `prefix`, `bin`, or `lib`, it gets paths pointing into zerobrew's Cellar instead of Homebrew's. This means we're running the exact same build instructions that Homebrew runs, just in our own prefix.

The `resource` and `patch` systems are also worth calling out because they required a two-phase design. Both are defined at the class level during formula loading (Ruby class body evaluation) but consumed at the instance level during `install`. So the shim collects patch and resource definitions into class-level storage via `inherited`, then patches are applied after `load` but before `install`, while resources are downloaded on-demand when the formula calls `resource("name").stage`.

The only known gap is formulas that rely on language-specific build pipelines like `std_pip_args` or `std_cargo_args`. Those are a long tail we can fill in incrementally as users hit them, since each one is just another method stub on the shim's `Formula` class. I think what's in place now will thankfully make that easier once we tackle it.

maria-rcks (approved):

zb_io/src/build/shim.rb
```diff
@@ -0,0 +1,605 @@
+require "fileutils"
```
MikeMcQuaid, line 1: @cachebag This file copy-pastes a lot of Homebrew's code without attribution. Please ensure you read our license and provide the appropriate attribution and meet those license terms.

PR #217, opened 2026-02-11
feat(zb_cli, zb_io): add `zb repair` command to fix orphaned kegs

AM8RDEV:

#

Adds a `zb repair` command that detects and re-registers kegs which exist in the cellar directory but are missing from the database. This handles cases where the database gets out of sync with the actual installed kegs (e.g., interrupted installs, manual file operations, or database corruption).

**Changes:**
- Add `Cellar::scan_kegs()` to walk the cellar directory and return all `(name, version)` pairs on disk
- Add `Installer::repair()` to compare on-disk kegs against the DB and register missing entries
- Wire up `zb repair` CLI command with user-friendly progress output

This is a follow-up to #207 — provides a recovery path for kegs that were installed but not tracked in the database.

### DIFF
--- a/zb_io/src/cellar/materialize.rs
+++ b/zb_io/src/cellar/materialize.rs
@@ -110,10 +110,9 @@ impl Cellar {
 
             let name = entry.file_name().to_string_lossy().to_string();
 
-            let versions =
-                std::fs::read_dir(entry.path()).map_err(|e| Error::StoreCorruption {
-                    message: format!("failed to read keg directory for {name}: {e}"),
-                })?;
+            let versions = std::fs::read_dir(entry.path()).map_err(|e| Error::StoreCorruption {
+                message: format!("failed to read keg directory for {name}: {e}"),
+            })?;
 
             for version_entry in versions {
                 let version_entry = version_entry.map_err(|e| Error::StoreCorruption {
--- a/zb_cli/src/commands/repair.rs
+++ b/zb_cli/src/commands/repair.rs
@@ -17,7 +17,12 @@ pub fn execute(installer: &mut zb_io::Installer) -> Result<(), zb_core::Error> {
             style(repaired.len()).green().bold()
         );
         for (name, version) in &repaired {
-            println!("    {} {} {}", style("✓").green(), name, style(version).dim());
+            println!(
+                "    {} {} {}",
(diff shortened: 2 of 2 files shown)


zb_io/src/installer/install.rs
```diff
@@ -520,6 +520,30 @@ impl Installer {
...
+    pub fn repair(&mut self) -> Result<Vec<(String, String)>, Error> {
+        let on_disk = self.cellar.scan_kegs()?;
+        let mut repaired = Vec::new();
+
+        for (name, version) in &on_disk {
+            if self.db.get_installed(name).is_some() {
+                continue;
+            }
+
+            let store_key = format!("repaired-{}-{}", name, version);
```
cachebag, line 536: This gets inserted into `store_refs` by `record_install`, but no corresponding store entry exists on disk. gc will try to remove a nonexistent entry. 

Just look up the actual store hash from the cellar contents.

zb_io/src/installer/install.rs
```diff
@@ -520,6 +520,30 @@ impl Installer {
...
+    /// pairs that were registered.
+    pub fn repair(&mut self) -> Result<Vec<(String, String)>, Error> {
+        let on_disk = self.cellar.scan_kegs()?;
+        let mut repaired = Vec::new();
+
+        for (name, version) in &on_disk {
+            if self.db.get_installed(name).is_some() {
+                continue;
+            }
+
```
cachebag, line 535: This checks by name only, so a keg on disk at version 2.0.0 will be skipped if the DB has version 1.0.0. Please compare (name, version) pairs instead.

zb_io/src/installer/install.rs
cachebag, line 1: We need tests for this.

zb_io/src/cellar/materialize.rs
cachebag, line 1: No tests?

zb_cli/src/cli.rs
```diff
@@ -92,6 +92,7 @@ pub enum Commands {
         formula: String,
     },
     Gc,
+    Repair,
```
cachebag, line 95: Nit: This new variant has no help text. I'd say that other variants meanings are implicit in their names but this one is ambiguous and requires clarity

PR #228, opened 2026-02-12
fix(zb_io): preserve execute bit when patching Mach-O binary strings

cachebag:

#

`patch_macho_binary_strings` writes patched binaries via a temp file (`fs::File::create` which defaults to `0644`), then renames it over the original. The rename replaces the inode, losing the original file's execute bit (`0755` becomes `0644`).

This PR restores the original file's permissions after the rename.

### DIFF
--- a/zb_io/src/extraction/patch/macos.rs
+++ b/zb_io/src/extraction/patch/macos.rs
@@ -182,6 +182,12 @@ fn patch_macho_binary_strings(path: &Path, new_prefix: &str) -> Result<(), Error
             message: format!("failed to rename temp file: {e}"),
         })?;
 
+        // Restore original permissions — fs::File::create uses 0644 by default,
+        // which drops the execute bit from patched binaries.
+        fs::set_permissions(path, metadata.permissions()).map_err(|e| Error::StoreCorruption {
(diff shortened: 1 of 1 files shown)


zb_io/src/extraction/patch/macos.rs
```diff
@@ -545,8 +537,39 @@ pub fn codesign_and_strip_xattrs(keg_path: &Path) -> Result<(), Error> {
 mod tests {
     use super::*;
     use std::fs;
+    use std::os::unix::fs::PermissionsExt;
     use tempfile::TempDir;
 
+    #[test]
+    fn test_patch_macho_preserves_execute_bit() {
+        let tmp = TempDir::new().unwrap();
```
UrsUbigfoot, line 545: If TempDir value is null or none, rust will panic. this will need to be validated before unwrap is used.

cachebag: `TempDir` returns a `Result`, not an `Option`. 

Even if it returns `Result`, using `unwrap()` in a unit test setup is entirely conventional. If a temp directory cannot be created, the test cannot proceed meaningfully. A panic is appropriate. There is no additional “validation” required before `unwrap()` in this context.

PR #231, opened 2026-02-13
feat(zb_io): resolve tap formulas from Formula, HomebrewFormula, and root

maria-rcks:
- [x] I have run all relevant linters for this PR.
- [x] I have read and will follow the contributing guidelines.

#

- [x] AI was used in this PR to generate code, documentation, or other changes, and I have explained _how_ it was used below. I understand that if I do not adhere to disclosing AI usage, my PR can be closed, even without explanation.
- [ ] I have not used AI for _any_ portion of this PR.

changes:
- adds UnsupportedFormula for formulas that where able to fetch but not install. `missing bottle`
- now we go through Formula, HomebrewFormula and then repo root to fetch .rb file.

> [!NOTE]
>  this pr is made to fix issues with tap `jandedobbeleer/oh-my-posh/oh-my-posh`, in another pr i'll try and add support for installing non-bottle formulas
> this changes only make it so zb finds the formula, but still cant install it

> AI-Assisted: planning / intial implementation made by gpt-5.3-codex, manually reviewed and tested

PR #232, opened 2026-02-13
feat(zb_io): support source-only tap formulas 

maria-rcks:

#

<!--
If you used AI assistance for any part of this pull request (code, documentation, or any other contributions), you must disclose this by checking the last box in this PR template and, if possible, add a note in your PR description specifying what was generated or significantly assisted by AI. This helps us ensure transparency and maintain a high standard for review. Thank you!
-->

<!--
Now you may use this space to discuss any other relevant information about this PR as well as link any related issues or PRs.
-->

   follow-up to #231, this one adds support for source-only tap formulas (no `bottle do`) making `jandedobbeleer/oh-my-posh/oh-my-posh` installation possible.                                             
                                                                                                                                        
   changes:                                                                                                                             
   - supports tap formulas that only define `url`/`sha256` (+ build deps), even when no bottle block exists                             
   - keeps tap ruby source location so source builds fetch the correct tap `.rb` file (not homebrew/core)                               
   - hardens tap ruby parsing to only read top-level formula fields (ignores nested `resource do` blocks for url/sha/deps)              
   - streams source build stdout/stderr live                                              
   - includes stderr tail in source-build error message for better debugging                                                            
   - adds shim support for `mv` with kwargs/options compatibility                                                                       
   - removes stale source-build keg output before build                 
                                                                                                                                        
   tests:                                                                                                                               
   - new tap parser tests for:                                                                                                          
     - source-only formulas without bottle blocks                                                                                       
     - nested resource block parsing safety (url/sha/deps)                                                                              
   - new api tests for:                                                                                                                 
     - source-only tap formula fetch path                                                                                               
     - encoded tap ruby source url handling                                                                                             
   - new build executor tests for:                                                                                                      
     - `mv` support in install flow                                                                                                     
     - stderr tail included on source-build failure                                                                                     
                         
                         
   > [!NOTE]
   > #231 needs to be merged first (this pr is built on top of i)                                                                                                                                                                 
                                                                                                                                        
   > AI-assisted: planning, implementation, and test drafting with gpt-5.3-codex; manually reviewed and validated.

### DIFF
--- a/zb_io/src/network/api.rs
+++ b/zb_io/src/network/api.rs
@@ -4,7 +4,50 @@ use crate::network::tap_formula::{parse_tap_formula_ref, parse_tap_formula_ruby}
 use futures_util::stream::{self, StreamExt};
 use zb_core::{Error, Formula};
 
-const TAP_RB_URL_PREFIX: &str = "tap-rb-url:";
+const HOMEBREW_CORE_RAW_BASE: &str =
+    "https://raw.githubusercontent.com/Homebrew/homebrew-core/main";
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum RubySourceLocator<'a> {
+    CoreRelativePath(&'a str),
+    AbsoluteUrl(&'a str),
+    TapEncodedUrl(&'a str),
+}
+
+impl<'a> RubySourceLocator<'a> {
+    const TAP_URL_PREFIX: &'static str = "tap-rb-url:";
+
+    fn parse(input: &'a str) -> Self {
+        if let Some(encoded_url) = input.strip_prefix(Self::TAP_URL_PREFIX) {
+            return Self::TapEncodedUrl(encoded_url);
+        }
+
+        if input.starts_with("https://") || input.starts_with("http://") {
+            return Self::AbsoluteUrl(input);
+        }
+
+        Self::CoreRelativePath(input)
+    }
+
+    fn source_id(self, original: &'a str) -> &'a str {
+        match self {
+            Self::CoreRelativePath(_) => original,
+            Self::AbsoluteUrl(url) => url,
+            Self::TapEncodedUrl(url) => url,
+        }
+    }
+
+    fn to_url(self) -> String {
+        match self {
+            Self::CoreRelativePath(path) => format!("{HOMEBREW_CORE_RAW_BASE}/{path}"),
+            Self::AbsoluteUrl(url) | Self::TapEncodedUrl(url) => url.to_string(),
+        }
+    }
+
+    fn encode_tap_url(url: &str) -> String {
+        format!("{}{}", Self::TAP_URL_PREFIX, url)
+    }
+}
 
 pub struct ApiClient {
     base_url: String,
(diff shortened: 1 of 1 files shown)


zb_io/src/network/tap_formula.rs
```diff
@@ -120,54 +144,168 @@ fn parse_revision(source: &str) -> Option<u32> {
...
+        *depth = depth.saturating_sub(1);
+        return;
+    }
+
+    *depth += DO_RE.find_iter(trimmed).count();
+    if KEYWORD_START_RE.is_match(trimmed) {
+        *depth += 1;
+    }
+}
+
```
cachebag, line 272: `DO_RE` is `\bdo\b` and it matches inside string literals... 

For instance: consider a formula with desc "A tool to do amazing things" increments depth with no matching end, so every subsequent top-level `depends_on`, `url`, and `sha256` line gets skipped. You need to anchor `do` to end-of-line patterns like `\bdo\s*$` or` `\bdo\s*\|`.

zb_io/src/network/tap_formula.rs
```diff
@@ -120,54 +144,168 @@ fn parse_revision(source: &str) -> Option<u32> {
...
+fn extract_formula_class_body(source: &str) -> Option<&str> {
```
cachebag, line 273: This has its own inline depth tracking that duplicates `update_depth` logic. If one gets fixed (such as for the do-in-strings bug above in `update_depth`), the other won't. It should call `update_depth` internally instead of reimplementing it.

zb_io/src/network/api.rs
```diff
@@ -1,8 +1,11 @@
 use crate::checksum::verify_sha256_bytes;
 use crate::network::cache::{ApiCache, CacheEntry};
 use crate::network::tap_formula::{parse_tap_formula_ref, parse_tap_formula_ruby};
+use futures_util::stream::{self, StreamExt};
 use zb_core::{Error, Formula};
 
+const TAP_RB_URL_PREFIX: &str = "tap-rb-url:";
```
cachebag: This is fragile for extensibility. If we add another source kind later, you're doing ad-hoc protocol parsing on a `String`. We'd want compile-time exhaustiveness checks here

zb_io/src/network/tap_formula.rs
```diff
@@ -120,54 +144,168 @@ fn parse_revision(source: &str) -> Option<u32> {
...
-    stable: &str,
-    revision: u32,
-) -> Result<Bottle, Error> {
-    let block = extract_bottle_block(source).ok_or_else(|| Error::MissingFormula {
-        name: format!(
-            "tap formula '{}' does not contain a bottle block",
-            spec.formula
-        ),
-    })?;
+fn parse_source_url(source: &str) -> Option<SourceUrl> {
```
cachebag: A formula with `url` but no `sha256` returns `Some(SourceUrl { checksum: None, .. })` and is accepted as installable. That means source downloads with no integrity verification. 

This should require a checksum and return `UnsupportedFormula` without one.

zb_io/src/installer/install.rs
```diff
@@ -574,6 +574,16 @@ impl Installer {
             }
         }
 
+        let keg_path = self.cellar.keg_path(formula_name, &version);
+        if keg_path.exists() {
+            fs::remove_dir_all(&keg_path).map_err(|e| Error::StoreCorruption {
```
cachebag: Doing this on the existing keg before the build starts means if the build fails, the previous working installation is gone with no rollback. I guess it's fine for cleaning up a partial previous build? But nothing gates it to that case. If someone reinstalls a source formula and the build fails, they've lost their working install. 

Either rename the old keg to a temp path and restore on failure, or check that the keg isn't currently recorded as installed before nuking it. I don't actually have a preference for what you do here.

PR #234, opened 2026-02-13
ci: run homebrew compataibility tests every sunday night

cachebag:

#

This workflow runs every Sunday night at 9PM ET and tests zb and our ruby shim to ensure compatibility with homebrew. If something breaks, an issue on the repo is filed with the version it broke on.

This provides early warning when Homebrew DSL changes break our shim compatibility.

### DIFF
--- a/.github/workflows/homebrew-compat.yml
+++ b/.github/workflows/homebrew-compat.yml
@@ -1,0 +1,201 @@
+name: Homebrew Compatibility
+
+on:
+  schedule:
+    # Run weekly on Sunday at 9 PM ET (1 AM UTC Monday)
+    - cron: '0 1 * * 1'
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'zb_io/src/build/shim.rb'
+      - '.github/workflows/homebrew-compat.yml'
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  test-compatibility:
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


.github/workflows/homebrew-compat.yml
```diff
@@ -0,0 +1,144 @@
+name: Homebrew Compatibility
+
+on:
```
maria-rcks, line 3: maybe add explicit permisssions:  so it checks if it can create issues:
```yaml
permissions:
  contents: read
  issues: write
```

.github/workflows/homebrew-compat.yml
```diff
@@ -0,0 +1,144 @@
...
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install Homebrew (Linux)
```
maria-rcks: is there a better way than doing a curl install? if not im totally fine with this approach but maybe something more stable is desirable.

.github/workflows/homebrew-compat.yml
```diff
@@ -0,0 +1,144 @@
...
+
+            **Action Items:**
+            - Review failed formulas
+            - Check for Homebrew DSL changes
+            - Update shim.rb if necessary
+            - Update compatibility documentation
+
+            **Workflow Run:** ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}`;
+
+            await github.rest.issues.create({
```
maria-rcks: although this runs weekly, but its also triggered all the time when pull requests are open (modifying homebrew compat that is) we can't be certain this issues will be dealt in time,

consider adding a check for if an existing issue has already been created

PR #252, opened 2026-02-23
feat(zb_io): ZEROBREW_API_URL, persistent API cache, contract tests

### LINKED ISSUE #259: Phase 0: API contract tests, ZEROBREW_API_URL support, and cache wiring
Parent: #258

## Reason

The Homebrew formulae API at `formulae.brew.sh` is an external dependency with no versioning guarantees. If the API schema changes (field renames, structural changes), zerobrew will silently break. Additionally, the existing `ApiCache` is created in-memory during each session — meaning every `zb install` re-fetches all metadata from scratch, wasting bandwidth and time. Finally, there's no way for users or CI to point zerobrew at an alternative API endpoint for testing or private taps.

## Impact

- **Regression detection**: Contract smoke tests catch upstream API changes before they break users
- **Performance**: Persistent SQLite cache means repeated installs skip redundant HTTP requests (ETag/Last-Modified conditional fetches already exist but the cache was ephemeral)
- **Testability**: `ZEROBREW_API_URL` lets CI and advanced users point at mock or alternative endpoints
- **Foundation**: Every subsequent phase (update, outdated, upgrade) depends on the cache being persistent and the API client being configurable

## Approach

1. **`ZEROBREW_API_URL` env var** in `ApiClient::new()` — validate URL scheme (http/https only), reject credentials in URL, fall back to default `formulae.brew.sh`
2. **Persistent cache wiring** in `create_installer()` — create `$ZEROBREW_ROOT/cache/api-cache.sqlite`, pass it to `ApiClient`
3. **Contract smoke tests** (`zb_io/tests/api_contract.rs`) — 3 `#[ignore]` tests that hit the real API to validate schema, parsing, and 404 behavior. Run on-demand with `cargo test --ignored`.

Files changed:
- `zb_io/src/network/api.rs` — `new() -> Result`, URL validation
- `zb_io/src/network/cache.rs` — `Debug` impl for `ApiCache`
- `zb_io/src/installer/install.rs` — cache directory + persistent cache wiring
- `zb_io/tests/api_contract.rs` — new contract test file

## Out of Scope / Next Opportunities

- **Cache TTL or expiry** — currently cache lives until `zb update` clears it; a time-based eviction policy could be added later
- **Cache size limits** — no pruning of old entries; could add LRU eviction if the SQLite file grows large
- **Multiple API endpoints** — `ZEROBREW_API_URL` supports one URL; multi-source federation is a future feature
- **Tap-specific API caching** — tap formulas use a different fetch path; unifying them under the cache is a follow-up

Relates to #258

aronchick:
## Summary

- Add `ZEROBREW_API_URL` env var support to `ApiClient::new()` with URL validation (scheme + credential checks)
- Wire persistent SQLite API cache into `create_installer()` at `$ZEROBREW_ROOT/cache/api-cache.sqlite`
- Add 3 contract smoke tests (`#[ignore]`) that validate the Homebrew API schema hasn't changed

## Test Plan

- [x] `cargo test --workspace` — all existing tests pass
- [x] `cargo test --ignored -p zb_io` — contract tests pass against live API
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean

Closes #259

---

- [x] I have read and will follow the [contributing guidelines](../blob/main/CONTRIBUTING.md).
- [x] AI was used in this PR. I used Claude Code to assist with implementation following the plan discussed in #258. All code was reviewed and tested by the PR author.

### DIFF
--- a/zb_io/src/installer/install.rs
+++ b/zb_io/src/installer/install.rs
@@ -1328,7 +1328,7 @@ mod tests {
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri());
+        let api_client = ApiClient::with_base_url(mock_server.uri()).unwrap();
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
         let cellar = Cellar::new(&root).unwrap();
@@ -1417,7 +1417,7 @@ mod tests {
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri());
+        let api_client = ApiClient::with_base_url(mock_server.uri()).unwrap();
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
         let cellar = Cellar::new(&root).unwrap();
@@ -1505,7 +1505,7 @@ mod tests {
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri());
+        let api_client = ApiClient::with_base_url(mock_server.uri()).unwrap();
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
         let cellar = Cellar::new(&root).unwrap();
@@ -1603,7 +1603,7 @@ mod tests {
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri());
+        let api_client = ApiClient::with_base_url(mock_server.uri()).unwrap();
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
         let cellar = Cellar::new(&root).unwrap();
@@ -1728,7 +1728,7 @@ mod tests {
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri());
+        let api_client = ApiClient::with_base_url(mock_server.uri()).unwrap();
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
         let cellar = Cellar::new(&root).unwrap();
@@ -1817,7 +1817,8 @@ end
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri())
+        let api_client = ApiClient::with_base_url(mock_server.uri())
+            .unwrap()
             .with_tap_raw_base_url(mock_server.uri());
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
@@ -1890,7 +1891,8 @@ end
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri())
+        let api_client = ApiClient::with_base_url(mock_server.uri())
+            .unwrap()
             .with_tap_raw_base_url(mock_server.uri());
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
@@ -1970,7 +1972,7 @@ end
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri());
+        let api_client = ApiClient::with_base_url(mock_server.uri()).unwrap();
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
         let cellar = Cellar::new(&root).unwrap();
@@ -2087,7 +2089,7 @@ end
         let prefix = tmp.path().join("homebrew");
         fs::create_dir_all(root.join("db")).unwrap();
 
-        let api_client = ApiClient::with_base_url_unchecked(mock_server.uri());
+        let api_client = ApiClient::with_base_url(mock_server.uri()).unwrap();
         let blob_cache = BlobCache::new(&root.join("cache")).unwrap();
         let store = Store::new(&root).unwrap();
         let cellar = Cellar::new(&root).unwrap();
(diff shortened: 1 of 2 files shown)


.gitignore
cachebag, line 1: Unneeded change. Why was this added?

aronchick: Good catch — removed. I had a local plans directory and included it by mistake. The updated commit no longer touches .gitignore.

zb_io/src/network/api.rs
```diff
@@ -58,8 +59,31 @@ pub struct ApiClient {
 }
 
 impl ApiClient {
-    pub fn new() -> Self {
-        Self::with_base_url("https://formulae.brew.sh/api/formula".to_string())
+    /// Create a new ApiClient. If `ZEROBREW_API_URL` is set, uses it as the
+    /// base URL (validated: http/https only, no embedded credentials).
+    pub fn new() -> Result<Self, Error> {
```
cachebag: Why did you change this but not update the call-sites?

zb_io/src/network/api.rs
cachebag, line 424: If someone has a malformed `ZEROBREW_API_URL` set in their environment, calling `ApiClient::default()` will panic instead of returning an error. This is a footgun.

zb_io/src/network/api.rs
```diff
@@ -397,7 +421,7 @@ impl ApiClient {
 
 impl Default for ApiClient {
     fn default() -> Self {
-        Self::new()
+        Self::new().expect("failed to create default ApiClient")
```
cachebag: If someone has a malformed `ZEROBREW_API_URL` set in their environment, calling `ApiClient::default()` will panic instead of returning an error. This is a footgun. What's your defense for having a default impl at all?

zb_io/src/network/api.rs
```diff
@@ -408,6 +432,39 @@ mod tests {
...
+        // Helper to test with a specific env var value
+        let test_url = |url: &str| -> Result<ApiClient, Error> {
+            // SAFETY: This test is single-threaded (#[test], not #[tokio::test])
+            // and no other code reads ZEROBREW_API_URL concurrently.
+            unsafe { std::env::set_var("ZEROBREW_API_URL", url) };
+            let result = ApiClient::new();
+            unsafe { std::env::remove_var("ZEROBREW_API_URL") };
+            result
+        };
+
```
cachebag: `cargo test` runs tests in parallel by default across the entire binary. If any other test in the same binary reads `ZEROBREW_API_URL` (or calls `ApiClient::new()/default())`, there's a race. This is quite fragile.

zb_io/src/network/api.rs
```diff
@@ -58,8 +59,31 @@ pub struct ApiClient {
...
+                ),
+            });
+        }
+        if !parsed.username().is_empty() || parsed.password().is_some() {
+            return Err(Error::InvalidArgument {
+                message: "ZEROBREW_API_URL must not contain credentials".to_string(),
+            });
+        }
+
+        Ok(Self::with_base_url(base_url))
```
cachebag: Why is this unchecked? i.e. why is there no validation

zb_io/src/network/api.rs
```diff
@@ -58,8 +59,31 @@ pub struct ApiClient {
...
+            return Err(Error::InvalidArgument {
+                message: format!(
+                    "ZEROBREW_API_URL must use http or https scheme, got: {}",
+                    parsed.scheme()
+                ),
+            });
+        }
+        if !parsed.username().is_empty() || parsed.password().is_some() {
+            return Err(Error::InvalidArgument {
+                message: "ZEROBREW_API_URL must not contain credentials".to_string(),
```
cachebag:
```suggestion
                message: "Bad ZEROBREW_API_URL configuration".to_string(),
                
```

Something like this would be better.

zb_io/src/installer/install.rs
```diff
@@ -1106,7 +1107,18 @@ pub fn create_installer(
         message: format!("failed to create db directory: {e}"),
     })?;
 
-    let api_client = ApiClient::new();
+    // Ensure cache directory exists before opening API cache SQLite
+    fs::create_dir_all(root.join("cache")).map_err(|e| Error::StoreCorruption {
+        message: format!("failed to create cache directory: {e}"),
+    })?;
+
+    // Wire persistent API cache for HTTP response caching (ETag/Last-Modified)
```
cachebag: Why is this comment here?

cachebag (changes requested):
There's several seemingly LLM generated comments. Please remove them if they're not explaining anything particularly interesting about the code.

aronchick:
Sorry about the LLM comments - i had it over comment since this is a public project. I've cleaned up a bunch.

Now just restate what the function name/signature already says (/// Clear the API cache..., /// Check if a
  specific installed package is outdated, /// Upgrade a single package atomically, etc.).
  
  Kept only the UpgradeResult variant docs since those explain non-obvious semantics like "package may be broken."

zb_io/src/network/api.rs
aronchick: Refactored. new() is now infallible again (returns Self, uses the hardcoded default URL). The env var reading moved out of the library entirely - create_installer() in install.rs now reads ZEROBREW_API_URL and passes it to with_base_url() which does validation. This way the library has no side effects from global state, and all call-sites that used ApiClient::new() work without changes.

zb_io/src/network/api.rs
aronchick: Fixed. new() no longer reads the env var, so Default just delegates to new() which always succeeds with the hardcoded Homebrew URL. No panic path.

zb_io/src/network/api.rs
aronchick: Now that new() is infallible, clippy actually requires a Default impl (clippy::new_without_default). It's just Self::new() with no env var reading, no validation, no error path — purely the hardcoded default URL. Safe to keep.

zb_io/src/network/api.rs
aronchick: Eliminated entirely. The unsafe { env::set_var } test is gone. Replaced with 5 direct unit tests that call with_base_url(url) with various inputs - no env var mutation, no race conditions, safe to run in parallel:

  - with_base_url_rejects_non_http_schemes
  - with_base_url_rejects_embedded_credentials
  - with_base_url_rejects_garbage
  - with_base_url_accepts_valid_https
  - with_base_url_accepts_valid_http

zb_io/src/network/api.rs
aronchick: Fixed. with_base_url() now validates (returns Result<Self, Error>). For internal test usage where we pass mock server URLs  (http://127.0.0.1:PORT), there's a #[cfg(test)] with_base_url_unchecked() that skips validation. Production code always goes through the validated path

zb_io/src/network/api.rs
aronchick: Done - changed the credential error message to "Bad ZEROBREW_API_URL configuration".

zb_io/src/installer/install.rs
aronchick: Removed. That was // Wire persistent API cache for HTTP response caching (ETag/Last-Modified) - just restating what the code does. Gone now, along with the other obvious comments.

cachebag (commented):
LGTM! Thanks.

PS We're gonna hold off on your other PR's regarding the new commands. We're still ironing out the design patterns we want internally.

I'll lyk if anything changes!

cachebag (approved):

aronchick:
Sorry about the previous flood. I hope you aren’t waving me off because of
the previous ones. I just really want to use this to replace home brew and
got excited to help

PR #266, opened 2026-02-23
feat(zb_cli): implement zb outdated command

aronchick:
- [X] I have read and will follow the [contributing guidelines](../blob/main/CONTRIBUTING.md).

# 

- Implement `zb outdated` CLI command with three output modes:
  - Default: `formula (installed → current)` with styled output
  - `--quiet`: formula names only (for piping to `xargs zb upgrade`)
  - `--verbose`: adds sha256 and build type details
  - `--json`: machine-readable JSON array
- Mutual exclusivity enforced via clap `conflicts_with_all`
- Switch `check_outdated()` from N individual `get_formula()` calls to a single bulk `get_all_formulas()` request (`formula.json` endpoint), with per-formula fallback only for tap formulas

### DIFF
--- a/zb_cli/src/commands/outdated.rs
+++ b/zb_cli/src/commands/outdated.rs
@@ -1,0 +1,61 @@
+use console::style;
+
+pub async fn execute(
+    installer: &mut zb_io::Installer,
+    quiet: bool,
+    verbose: bool,
+    json: bool,
+) -> Result<(), zb_core::Error> {
+    let (outdated, warnings) = installer.check_outdated().await?;
+
+    // Warnings always go to stderr (never pollute stdout, especially in --json mode)
+    for warning in &warnings {
+        eprintln!("{} {}", style("Warning:").yellow().bold(), warning);
+    }
+
+    if json {
+        let json_output: Vec<serde_json::Value> = outdated
+            .iter()
+            .map(|pkg| {
+                serde_json::json!({
+                    "name": pkg.name,
+                    "installed_versions": [pkg.installed_version],
+                    "current_version": pkg.current_version,
+                })
+            })
+            .collect();
+        println!("{}", serde_json::to_string_pretty(&json_output).unwrap());
+        return Ok(());
+    }
+
+    if outdated.is_empty() {
+        if !quiet {
+            println!(
+                "{} All packages are up to date.",
+                style("==>").cyan().bold()
+            );
+        }
+        return Ok(());
+    }
+
+    for pkg in &outdated {
+        if quiet {
+            println!("{}", pkg.name);
+        } else if verbose {
+            println!(
+                "{} {} {} {}",
+                pkg.name,
+                style(&pkg.installed_version).red(),
+                style("→").dim(),
+                style(&pkg.current_version).green(),
+            );
+        } else {
+            println!(
+                "{} ({}) < {}",
+                pkg.name, pkg.installed_version, pkg.current_version
+            );
+        }
+    }
+
+    Ok(())
+}
--- a/zb_cli/src/bin/zb.rs
+++ b/zb_cli/src/bin/zb.rs
@@ -64,6 +64,11 @@ async fn run(cli: Cli) -> Result<(), zb_core::Error> {
         Commands::Info { formula } => commands::info::execute(&mut installer, formula),
         Commands::Gc => commands::gc::execute(&mut installer),
         Commands::Update => commands::update::execute(&mut installer),
+        Commands::Outdated {
+            quiet,
+            verbose,
+            json,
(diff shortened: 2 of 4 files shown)


zb_io/src/installer/install.rs
```diff
@@ -70,6 +86,125 @@ impl Installer {
...
+
+        let api = &self.api_client;
+        let fetches: Vec<_> = futures_util::stream::iter(installed.iter().map(|keg| {
+            let name = keg.name.clone();
+            async move {
+                let result = api.get_formula(&name).await;
+                (name, result)
+            }
+        }))
+        .buffer_unordered(10)
```
cachebag: Consider what happens when someone has 100+ formulae; this fires one HTTP request per package.

zb_cli/src/commands/outdated.rs
```diff
@@ -0,0 +1,63 @@
...
+            .iter()
+            .map(|pkg| {
+                serde_json::json!({
+                    "name": pkg.name,
+                    "installed_versions": [pkg.installed_version],
+                    "current_version": pkg.current_version,
+                    "pinned": false,
+                    "pinned_version": null,
+                })
+            })
```
cachebag, line 25: zerobrew doesn't support pinning right now. Please drop these fields

zb_cli/src/commands/update.rs
```diff
@@ -0,0 +1,16 @@
+use console::style;
+
+pub fn execute(installer: &mut zb_io::Installer) -> Result<(), zb_core::Error> {
```
cachebag, line 3: `zb update` just clears the cache. the actual re-fetch happens lazily on the next command. This is different from `brew update` which actively syncs metadata. Worth adding a `println!` hint like "Run zb outdated to check for updates." so users aren't confused.

cachebag:
Please ensure you also resolve any conflicts @aronchick

.gitignore
cachebag, line 1: Once again, let's remove this.

zb_cli/src/commands/outdated.rs
aronchick: Removed

zb_cli/src/commands/update.rs
aronchick: you got it!

zb_io/src/installer/install.rs
aronchick: lol, wasn't even aware brew had a bulk endpoint. changed :)

aronchick:
> I don't really see the point in this commit. Addressing review feedback should not be a part of this history. Just make the changes and amend your initial commits. This should ideally all be done through an interactive rebase session where each respective change placed in the commit it was originally in.

No worries! On other projects i work on, they want to see every commit. My apologies, will ONLY do squashes and rebases.

zb_cli/src/commands/outdated.rs
aronchick: or did you mean the whole struct? Sorry, wasn't sure.

zb_io/src/network/api.rs
```diff
@@ -288,6 +298,75 @@ impl ApiClient {
         Ok(formula)
     }
 
+    pub async fn get_all_formulas(&self) -> Result<Vec<Formula>, Error> {
```
cachebag: This is basically just the same as `get_formula()`. Extract it into a helper.

zb_io/src/installer/install.rs
```diff
@@ -71,6 +86,121 @@ impl Installer {
...
+            let formula = if is_tap || !bulk_map.contains_key(keg.name.as_str()) {
+                match self.api_client.get_formula(&keg.name).await {
+                    Ok(f) => f,
+                    Err(e) => {
+                        warnings.push(format!("{}: {}", keg.name, e));
+                        continue;
+                    }
+                }
+            } else {
+                (*bulk_map[keg.name.as_str()]).clone()
```
cachebag: This clones each matched `Formula` and since the bulk response can contain 6000+ formulas, the map + clone approach means all of them are deserialized and held in memory even if the user only has 10 installed. 

Just deserialize this to `HashMap<String, serde_json::Value>` and only fully parse the ones that match installed packages.

zb_io/src/network/api.rs
aronchick: done!

zb_io/src/installer/install.rs
aronchick: I have now addressed this in a a perf commit (now squashed into this one). The original version made N individual get_formula() calls. The current code fetches the bulk endpoint once, parses to Vec<serde_json::Value>, and only deserializes matching entries into Formula.

PR #274, opened 2026-02-26
refactor(zb_cli): centralize CLI UI output

maria-rcks:

#

refactor to reduce printing logic.

- Added shared `Ui` layer in `zb_cli` (theme/symbols/styles + prompt helpers)
- Migrated command output to shared UI:
  - install
  - migrate
  - reset
  - uninstall
  - init
- Kept existing user-facing install UX mostly unchanged

> AI-Assited by gpt 5.3-codex for planning & implementation, manually reviewed and tested

### DIFF
--- a/zb_cli/src/commands/install.rs
+++ b/zb_cli/src/commands/install.rs
@@ -5,7 +5,7 @@ use std::sync::{Arc, Mutex};
 use std::time::Instant;
 use zb_io::{InstallProgress, ProgressCallback};
 
-use crate::ui::Ui;
+use crate::ui::StdUi;
 use crate::utils::{normalize_formula_name, suggest_homebrew};
 
 pub async fn execute(
@@ -13,9 +13,9 @@ pub async fn execute(
     formulas: Vec<String>,
     no_link: bool,
     build_from_source: bool,
+    ui: &mut StdUi,
 ) -> Result<(), zb_core::Error> {
     let start = Instant::now();
-    let mut ui = Ui::new();
     ui.heading(format!(
         "Installing {}...",
         style(formulas.join(", ")).bold()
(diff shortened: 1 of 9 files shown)


maria-rcks:
there

cachebag:
> there

I meant two separate pull requests.

>Could you split these into two PRs? The UI changes and tracing are unrelated to eachother.

maria-rcks:
I removed the commits for the logging, is just now the ui part, i'll create the other pr ina  second

zb_cli/src/ui.rs
```diff
@@ -0,0 +1,296 @@
...
+        let label = self
+            .theme
+            .styles
+            .note_label
+            .apply_to(self.theme.symbols.note_label)
+            .to_string();
+        writeln!(self.out, "{label} {message}")
+    }
+
+    pub fn info(&mut self, message: impl Display) -> io::Result<()> {
```
cachebag, line 130: For lines like `println!("No Homebrew packages installed.")`, we used to just print bare text. Now we prepend "Info" to everything.

Please use something like plain `writeln!` instead (e.g. `ui.println()`) for messages that were previously bare text. I'm assuming you didn't mean to make this regression

zb_cli/src/commands/install.rs
```diff
@@ -186,10 +187,8 @@ pub async fn execute(
             Ok(r) => r,
             Err(ref e @ zb_core::Error::LinkConflict { ref conflicts }) => {
                 eprintln!();
-                eprintln!(
-                    "{} The link step did not complete successfully.",
-                    style("Error:").red().bold()
-                );
+                ui.error("The link step did not complete successfully.")
+                    .map_err(ui_error)?;
```
cachebag, line 191: Mixing this with `eprintln!()` everywhere else defeats the purpose of this PR. Let's fully migrate these

zb_cli/src/ui.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct UiTheme {
+    pub styles: UiStyles,
+    pub symbols: UiSymbols,
+}
+
+pub struct Ui<O: Write, E: Write> {
```
cachebag, line 75: Every command does `let mut ui = Ui::new()`. `init.rs` even creates two separate instances from what I see.

Since `Ui` owns `stdout`/`stderr` handles, this works today but makes it impossible to later add state (quiet mode, verbosity level) without threading a single `Ui` through. Just create this once at the top level and pass it down.

zb_cli/src/commands/install.rs
```diff
@@ -218,23 +217,29 @@ pub async fn execute(
...
         style(installed_count).green().bold(),
         elapsed.as_secs_f64()
-    );
+    ))
+    .map_err(ui_error)?;
 
     Ok(())
 }
+
+fn ui_error(err: std::io::Error) -> zb_core::Error {
```
cachebag, line 245: Just use `Error::FileError`, this makes no sense. An `stdout` write failure isn't store corruption.

zb_cli/src/commands/install.rs
cachebag: Oops, ignore this

PR #292, opened 2026-03-12
refactor: eliminate unwraps, reduce allocations, decompose install path

cachebag:

#

This pass focuses on a more mechanical cleanup across all three crates; including error handling, unnecessary allocations, and structural issues that accumulated from our initial implementation.

- Replaced all production `unwrap()` calls with `?` propagation. Notably: `init.rs` silently fell back to `"root"` before `sudo chown -R`, and the download module had 5 bare unwraps on bearer token header construction.

- Added closure-returning helpers on `Error` (`::store`, `::network`, `::file`, `::exec`) and applied across ~80 `map_err` sites in 12 files. Fixed `homebrew.rs` returning `Result<_, String>` and `BlobWriter` misclassifying filesystem errors as `NetworkFailure`.

- Rewrote dependency resolver from `BTreeMap<String, BTreeSet<String>>` with per-node clones to index-based toposort -- names interned to `usize` at the boundary, strings only touched on entry/exit.

- Replaced hand-rolled `pid+thread_id` temp file naming and TOCTOU-racy commit logic in blob/store with `tempfile::NamedTempFile` and `tempdir_in()`. Atomic `persist()`, auto-cleanup on drop.

- Extracted `process_bottle_item` and `record_linked_files` from the ~240-line inline loop body in `execute_with_progress`, replacing `match`/`continue`/`error = Some(e)` with `Result` propagation.

- `ApiClient` formula candidate cache changed from cloning `Vec<String>` to `Arc<[String]>`. `categorize_packages` takes `Vec` by value instead of borrowing and cloning every element.

- `InitError` implements `Display` and `From<std::io::Error>`, eliminating the `io_to_init_error` helper and its boilerplate.

This is more of a set of surface level hygiene patches that are jumpstarting a broader plan I have (see #293) to address some deeper architectural problems that pertain to the long term reliability of zerobrew. More PRs coming soon.

### DIFF
--- a/zb_core/src/errors.rs
+++ b/zb_core/src/errors.rs
@@ -77,30 +77,23 @@ impl fmt::Display for Error {
 
 impl std::error::Error for Error {}
 
-impl Error {
-    pub fn store<E: fmt::Display>(ctx: &str) -> impl FnOnce(E) -> Self + '_ {
-        move |err| Self::StoreCorruption {
-            message: format!("{ctx}: {err}"),
+macro_rules! error_helpers {
+    ($($fn_name:ident => $variant:ident),* $(,)?) => {
+        impl Error {
+            $(
+                pub fn $fn_name<E: fmt::Display>(ctx: &str) -> impl FnOnce(E) -> Self + '_ {
+                    move |err| Self::$variant { message: format!("{ctx}: {err}") }
+                }
+            )*
         }
-    }
+    };
+}
 
-    pub fn network<E: fmt::Display>(ctx: &str) -> impl FnOnce(E) -> Self + '_ {
-        move |err| Self::NetworkFailure {
-            message: format!("{ctx}: {err}"),
-        }
-    }
-
-    pub fn file<E: fmt::Display>(ctx: &str) -> impl FnOnce(E) -> Self + '_ {
-        move |err| Self::FileError {
-            message: format!("{ctx}: {err}"),
-        }
-    }
-
-    pub fn exec<E: fmt::Display>(ctx: &str) -> impl FnOnce(E) -> Self + '_ {
-        move |err| Self::ExecutionError {
-            message: format!("{ctx}: {err}"),
-        }
-    }
+error_helpers! {
+    store   => StoreCorruption,
+    network => NetworkFailure,
+    file    => FileError,
+    exec    => ExecutionError,
 }
 
 #[cfg(test)]
--- a/zb_io/src/installer/homebrew.rs
+++ b/zb_io/src/installer/homebrew.rs
@@ -97,12 +97,9 @@ pub fn get_homebrew_packages() -> Result<HomebrewMigrationPackages, Error> {
         .map_err(Error::exec("failed to run 'brew info'"))?;
 
     if !formulas_output.status.success() {
-        return Err(Error::ExecutionError {
-            message: format!(
-                "brew info failed: {}",
-                String::from_utf8_lossy(&formulas_output.stderr)
-            ),
-        });
+        return Err((Error::exec("brew info failed"))(String::from_utf8_lossy(
+            &formulas_output.stderr,
+        )));
     }
 
     let formulas_json: serde_json::Value = serde_json::from_slice(&formulas_output.stdout)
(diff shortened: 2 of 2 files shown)


zb_cli/src/commands/migrate.rs
```diff
@@ -11,14 +11,7 @@ pub async fn execute(
...
 
-    let packages = match zb_io::get_homebrew_packages() {
-        Ok(pkgs) => pkgs,
-        Err(e) => {
-            return Err(zb_core::Error::StoreCorruption {
-                message: format!("Failed to get Homebrew packages: {}", e),
-            });
-        }
-    };
+    let packages = zb_io::get_homebrew_packages()?;
```
lucasgelfond, line 14: nice

zb_cli/src/init.rs
```diff
@@ -10,6 +10,20 @@ pub enum InitError {
     Message(String),
 }
 
+impl std::fmt::Display for InitError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            InitError::Message(msg) => f.write_str(msg),
+        }
+    }
```
lucasgelfond, line 18: i haven't looked at this code recently but can't we usually derive string traits like this?

lucasgelfond: ok if not

zb_core/src/errors.rs
```diff
@@ -77,6 +77,32 @@ impl fmt::Display for Error {
...
+            message: format!("{ctx}: {err}"),
+        }
+    }
+
+    pub fn exec<E: fmt::Display>(ctx: &str) -> impl FnOnce(E) -> Self + '_ {
+        move |err| Self::ExecutionError {
+            message: format!("{ctx}: {err}"),
+        }
+    }
+}
```
lucasgelfond: same here / ideally DRY but this looks fine as is if that's cumbersome. i do wonder how many of these fns we can generate with macros though etc

zb_io/src/build/executor.rs
```diff
@@ -42,15 +42,11 @@ impl BuildExecutor {
         let shim_path = work_dir.join("zerobrew_shim.rb");
         fs::write(&shim_path, SHIM_RUBY)
             .await
-            .map_err(|e| Error::FileError {
-                message: format!("failed to write ruby shim: {e}"),
-            })?;
+            .map_err(Error::file("failed to write ruby shim"))?;
```
lucasgelfond, line 45: nice / these are nice

zb_io/src/installer/homebrew.rs
```diff
@@ -88,43 +88,50 @@ pub fn categorize_packages(packages: &[HomebrewPackage]) -> HomebrewMigrationPac
...
 /// Only formulas from `homebrew/core` can be migrated to zerobrew.
 /// Formulas from other taps and all casks are collected separately.
-pub fn get_homebrew_packages() -> Result<HomebrewMigrationPackages, String> {
-    // Get installed formulas
+pub fn get_homebrew_packages() -> Result<HomebrewMigrationPackages, zb_core::Error> {
     let formulas_output = Command::new("brew")
         .args(["info", "--json=v1", "--installed"])
         .output()
-        .map_err(|e| format!("Failed to run 'brew info': {}", e))?;
+        .map_err(|e| zb_core::Error::ExecutionError {
```
lucasgelfond: would love to sync this pattern up with the others

zb_io/src/installer/homebrew.rs
```diff
@@ -88,43 +88,50 @@ pub fn categorize_packages(packages: &[HomebrewPackage]) -> HomebrewMigrationPac
...
     if !formulas_output.status.success() {
-        return Err(format!(
-            "brew info failed: {}",
-            String::from_utf8_lossy(&formulas_output.stderr)
-        ));
+        return Err(zb_core::Error::ExecutionError {
+            message: format!(
+                "brew info failed: {}",
+                String::from_utf8_lossy(&formulas_output.stderr)
+            ),
```
lucasgelfond: same here

lucasgelfond (approved):
LGTM

zb_cli/src/init.rs
cachebag: You can't derive `Display`. rust doesn't know how a type should be formatted for user facing output

zb_core/src/errors.rs
cachebag: wonderful, yes a macro would be great here. thank you!

PR #305, opened 2026-03-13
fix(zb_io): added SQLite schema versioning with sequential migrations and downgrade protection

### LINKED ISSUE #295: add sqlite schema versioning and migration support
`init_schema` uses `CREATE TABLE IF NOT EXISTS` with no version tracking. Any future schema change silently fails on existing databases. No migration path.

Add a `schema_version` pragma or metadata table. Check on open, run migrations sequentially, reject downgrades. Needs to happen before any schema changes ship.

Part of #293

cachebag:

#

Closes #295 

Adds SQLite schema versioning for both the main database and API cache database using `PRAGMA user_version` to track versions. The migration system runs migrations sequentially from the current version to the target version, rejects downgrades by returning an error when the database version exceeds the code version, and preserves existing data through idempotent migrations.

### DIFF
--- a/zb_io/src/storage/db.rs
+++ b/zb_io/src/storage/db.rs
@@ -17,24 +17,71 @@ pub struct InstalledKeg {
 }
 
 impl Database {
+    const SCHEMA_VERSION: u32 = 1;
+
     pub fn open(path: &Path) -> Result<Self, Error> {
         let conn = Connection::open(path).map_err(Error::store("failed to open database"))?;
-
-        Self::init_schema(&conn)?;
-
+        Self::migrate(&conn)?;
         Ok(Self { conn })
     }
 
     pub fn in_memory() -> Result<Self, Error> {
         let conn =
             Connection::open_in_memory().map_err(Error::store("failed to open in-memory db"))?;
-
-        Self::init_schema(&conn)?;
-
+        Self::migrate(&conn)?;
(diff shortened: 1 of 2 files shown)


zb_io/src/storage/db.rs
```diff
@@ -471,4 +518,57 @@ mod tests {
...
+        conn.execute_batch(
+            "CREATE TABLE installed_kegs (
+                name TEXT PRIMARY KEY,
+                version TEXT NOT NULL,
+                store_key TEXT NOT NULL,
+                installed_at INTEGER NOT NULL
+            );
+            INSERT INTO installed_kegs VALUES ('test', '1.0.0', 'key123', 1234567890);",
+        )
+        .expect("failed to create old schema");
```
lucasgelfond, line 560: do we want an ORM? or too heavy?

cachebag:
> only q would be at what point should we add an ORM. my general intuition is: sooner rather than later! but cool with this for now

we're using SQLite as our user db rn- Diesel and SQLx adds many deps and increases the compile time in a pretty significant way.

i also feel like the DB is a simple ledger for installs/uninstalls and refcounts, thus it’s seems unlikely to turn into a complex domain model. but i get the vision; mainly im thinking about when/where to over-engineer and be confident in our decisions and when to just take the simplest path forward.

PR #311, opened 2026-03-15
fix(zb_io, zb_cli): validate root/prefix paths before passing to sudo

### LINKED ISSUE #296: validate root/prefix paths before passing to sudo
`init.rs` passes `root` and `prefix` to `sudo mkdir`/`sudo chown` via `to_string_lossy()` with zero validation. Uses `Command::new` so no shell injection, but CLI args going straight to privileged operations should be sanitized -- no control characters, no `..` traversal, reasonable length.

Part of #293

cachebag:

#

We now call `validate_privileged_path` on root and prefix at the top of `run_init` so `mkdir`/`chown` never receives unsanitized CLI input. `validate_privileged_path` was created to reject control characters, '..' traversal, and excessive length before paths reach sudo commands. 

Closes #296

### DIFF
--- a/zb_io/src/path.rs
+++ b/zb_io/src/path.rs
@@ -22,6 +22,12 @@ pub fn validate_privileged_path(path: &Path) -> Result<(), Error> {
         });
     }
 
+    if path_str.starts_with('-') {
+        return Err(Error::InvalidArgument {
+            message: format!("path starts with '-': {}", path.display()),
+        });
+    }
+
     for component in path.components() {
         if matches!(component, Component::ParentDir) {
             return Err(Error::InvalidArgument {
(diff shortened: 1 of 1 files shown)


zb_io/src/path.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    for component in path.components() {
+        if matches!(component, Component::ParentDir) {
+            return Err(Error::InvalidArgument {
+                message: format!("path contains '..' traversal: {}", path.display()),
+            });
+        }
+    }
+
+    Ok(())
```
maria-rcks, line 39: this still allows paths starting with `-`, so sudo can still parse the path as an option.
please reject leading-dash paths here or add -- before every user-controlled path

PR #314, opened 2026-03-17
feat(zb): `zb doctor` + `--repair` for state diagnosis and repair

### LINKED ISSUE #301: add zb doctor or zb repair for state recovery
No way to detect or repair: orphaned store refs, missing keg_files entries, inconsistent refcounts, partially materialized kegs. Only recovery is full reinstall.

A `zb doctor` that validates DB consistency, checks store/cellar agreement, and cleans up orphaned state.

Part of #293

cachebag:

#

Closes #301 

This PR adds `zb doctor` to detect and repair inconsistent state

Walks the cellar, store, DB, and prefix symlinks to find orphaned kegs, missing cellar directories, stale/mismatched store refcounts, orphaned store entries, broken symlinks, and dangling keg_files records. Diagnosis is read-only by default; `-repair` fixes everything it can; i.e. it removes orphans, prunes stale DB rows, recomputes refcounts from installed_kegs, and cleans broken links.

New introspection APIs on Cellar, Store, Linker, and Database provide the building blocks. No changes to existing install/uninstall behavior.

### DIFF
--- a/zb_io/src/installer/install/doctor.rs
+++ b/zb_io/src/installer/install/doctor.rs
@@ -100,7 +100,9 @@ impl Installer {
         };
 
         for entry in &disk_store_entries {
-            if !store_keys_in_db.contains(entry.as_str()) {
+            if !store_keys_in_db.contains(entry.as_str())
+                && !store_keys_used.contains_key(entry.as_str())
+            {
                 report.orphaned_store_entries.push(entry.clone());
             }
         }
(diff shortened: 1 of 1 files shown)


zb_io/src/installer/install/doctor.rs
```diff
@@ -0,0 +1,220 @@
...
+        let store_keys_used: HashMap<&str, i64> = {
+            let mut map = HashMap::new();
+            for keg in &installed {
+                *map.entry(keg.store_key.as_str()).or_insert(0) += 1;
+            }
+            map
+        };
+
+        for entry in &disk_store_entries {
+            if !store_keys_in_db.contains(entry.as_str()) {
```
maria-rcks: if a package still exists in installed_kegs but its store_refs row is missing, this store entry gets marked as orphaned and --repair will delete it even though it’s still in use.

we should also check whether this key is still referenced by any installed keg before calling it orphaned.

zb_io/src/installer/install/doctor.rs
```diff
@@ -0,0 +1,220 @@
...
+                    on_disk,
+                    referenced_by_any_keg: actual_count > 0,
+                });
+            }
+        }
+
+        for keg in &installed {
+            let token = formula_token(&keg.name);
+            let keg_path = self.cellar.keg_path(token, &keg.version);
+            if !keg_path.exists() {
```
maria-rcks: this skips broken-link detection for missing kegs. If the cellar path is already gone, the broken symlinks in the prefix won’t be found or cleaned up.

we probably need to use keg_files (or scan the prefix directly) instead of only walking the current keg path.

PR #332, opened 2026-03-31
fix(zb_io): resolve formula aliases and oldnames on 404

### LINKED ISSUE #326: zb can't install pkg-config
```bash
# ~/.local/bin/zb install pkg-config
==> Installing pkg-config...

Hint: Formula 'pkg-config' was not found. Did you mean:
      pkg-config
      pkgconfig
      pkgconf

      Try installing the closest match with zerobrew:
      zb install pkg-config

error: missing formula 'pkg-config'
```
But `pkgconf` work:
```bash
# ~/.local/bin/zb install pkgconf
==> Installing pkgconf...
==> Resolving dependencies (1 packages)...
    • pkgconf 2.5.1
==> Downloading and installing formulas...
    pkgconf          ✓ installed
==> Installed 1 packages in 0.98s
```

zb version: 0.2.1

pavvann:

#

Fixes #326

zb install pkg-config fails with "missing formula" because pkg-config is an alias for pkgconf in Homebrew's formula data — but zerobrew only used aliases for fuzzy suggestions, not for actual lookups.

When get_formula receives a 404, it now checks the bulk formula index for a matching alias or oldname and retries the fetch with the canonical name. The alias map is lazily built from the same bulk JSON already used for suggestions and cached in memory for the lifetime of the client.

### DIFF
--- a/zb_io/src/network/api.rs
+++ b/zb_io/src/network/api.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 
 use crate::checksum::verify_sha256_bytes;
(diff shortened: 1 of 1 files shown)


cachebag (approved):
LGTM. Thanks!

PR #333, opened 2026-03-31
feat(zb_io): make migration install only leaves

### LINKED ISSUE #187: Migration should use leaves
Instead of reinstalling every package individually during migration, `zb` should install the leaves (brew leaves) instead.

ordinary-hacker:

#

Resolves #187 

Now migration will only explicitly migrate Homebrew formulas which are _leaves_ (rather than all of them). Which is more efficient as there's no need to reinstall dependencies that will be installed by these _leaves_ anyways.

I only added a simple unit test to test the function to parse the output from `brew leaves` (which just checks that the _leaves_ are extracted correctly from the output), similarly to what other functions in the same module do.

Note there could, theoretically, be an edge case related to `ARG_MAX` limits, as all found leaves are passed to `brew info` as arguments at once. However, in most modern systems with more than 1 MB of `ARG_MAX` this becomes completely negligible, and even in extremely constrained systems with 128KB of `ARG_MAX` or less, thousands of packages would need to be installed to even overflow.

### DIFF
--- a/zb_io/src/installer/homebrew.rs
+++ b/zb_io/src/installer/homebrew.rs
@@ -45,6 +45,16 @@ pub fn parse_formulas_from_json(json: &serde_json::Value) -> Vec<HomebrewPackage
     packages
 }
 
+/// Parse Homebrew leaves from plain text output of `brew leaves`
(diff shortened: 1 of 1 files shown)


cachebag (approved):
LGTM

Issue #286, opened 2026-03-08
Intel Mac: Mach-O patching skipped when zerobrew prefix is longer than Homebrew prefix


jonirrings:
Hello,
I picked my old mbp with intel CPU to try zerobrew, and found the `gpgconf` in `GnuPG x86_64 build` with hardcoded  path:
```bash
strings `which gpgconf` | rg -i cellar
/usr/local/Cellar/gnupg/2.4.9/bin
/usr/local/Cellar/gnupg/2.4.9/libexec
/usr/local/Cellar/gnupg/2.4.9/lib/gnupg
/usr/local/Cellar/gnupg/2.4.9/share/gnupg
/usr/local/Cellar/gnupg/2.4.9/share/locale
```
and failed to start
```bash
gpgconf --launch gpg-agent
gpgconf: error running '/usr/local/Cellar/gnupg/2.4.9/bin/gpg-agent': probably not installed
gpgconf: Configuration file of component GPG Agent is broken
gpgconf: Note: Use the command "gpg-agent --gpgconf-test" to get details.
```

it there any way to fix this?

cachebag:
So this one's messy and has been haunting me for a while now. 

The Mach-O binary patcher replaces hardcoded Homebrew prefixes in-place (basically overwriting bytes directly in the binary). This requires `new_prefix.len() <= old_prefix.len()`. On Intel Mac, bottles embed `/usr/local` (which is 10 chars) but zerobrew's prefix is `/opt/zerobrew` (13 chars, bad news bears), so the patcher skips replacement entirely. 

All Intel bottles with hardcoded `/usr/local/Cellar/...` paths in their binary data sections are affected, not just gnupg (hence the reframing of the issue).

Load commands (shown by `otool -L`) are handled separately by `install_name_tool`, which can resize paths. But paths embedded as raw strings in the binary (like `gpgconf`'s component paths) cannot be grown without restructuring the Mach-O, which is very non-trivial.

For now we've added a warning so users see when this skip occurs. A real fix would require either a shorter prefix on Intel or Mach-O string table rewriting. Will use this issue for any updates and progress (or lack thereof).

fu050409:
@cachebag Hi, I don't know much about Mach-O. Will this issue affect Mac devices with aarch64 M-series chips?

cachebag:
> [@cachebag](https://github.com/cachebag) Hi, I don't know much about Mach-O. Will this issue affect Mac devices with aarch64 M-series chips?

[...]

PR #344, opened 2026-04-30
fix: skip binary patching when new prefix is longer instead of erroring

### LINKED ISSUE #342: Error: store corruption: zerobrew prefix "/opt/zerobrew/prefix" (20 bytes) is longer than "/opt/homebrew" (13 bytes).
╰─ x zb install aria2
==> Installing aria2...
==> Resolving dependencies (9 packages)...
    c-ares 1.34.6
    ca-certificates 2026-03-19
    libunistring 1.4.2
    gettext 1.0
    openssl@3 3.6.2
    libssh2 1.11.1
    readline 8.3.3
    sqlite 3.53.0
    aria2 1.37.0
==> Downloading and installing...
    readline         ✓ installed
    sqlite           ✓ installed
    aria2            ⠏ unpacking...
    gettext          ⠏ unpacking...
    ca-certificates  ✓ installed
    c-ares           ✓ installed
    openssl@3        ⠏ unpacking...
    libssh2          ✓ installed
    libunistring     ✓ installed
Note: This package can't be installed with zerobrew.
      Error: store corruption: zerobrew prefix "/opt/zerobrew/prefix" (20 bytes) is longer than "/opt/homebrew" (13 bytes). On macOS, Mach-O binaries cannot be expanded in-place, so path-sensitive formulae (e.g. git) will not work. Use a prefix of at most 13 characters, e.g. /opt/zerobrew. Re-init with: zb init <root> /opt/zerobrew (or your chosen short prefix), then reinstall.

      Try installing with Homebrew instead:
      brew install aria2

error: store corruption: zerobrew prefix "/opt/zerobrew/prefix" (20 bytes) is longer than "/opt/homebrew" (13 bytes). On macOS, Mach-O binaries cannot be expanded in-place, so path-sensitive formulae (e.g. git) will not work. Use a prefix of at most 13 characters, e.g. /opt/zerobrew. Re-init with: zb init <root> /opt/zerobrew (or your chosen short prefix), then reinstall.

KuaaMU:
## Problem

When zerobrew's prefix (`/opt/zerobrew/prefix`, 20 bytes) is longer than Homebrew's prefix (`/opt/homebrew`, 13 bytes), `patch_macho_binary_strings()` returns a fatal `StoreCorruption` error. This blocks installation of any package that has binaries referencing the Homebrew prefix.

## Root Cause

The function detects that in-place binary patching is impossible (can't expand shorter paths in Mach-O), then scans for old paths and returns an error if found. But this is wrong — the correct behavior is to **skip** (continue), because:

1. `install_name_tool` handles load command changes regardless of path length
2. Text files are patched separately in the text-file pass
3. The existing test `test_patch_macho_skips_when_new_prefix_longer` already expects `Ok(())`
4. CHANGELOG for PR #227 states: "Skip patching when new prefix is longer than old"

## Fix

Remove the error-returning logic (17 lines) and simplify to just `continue`. The comment is updated to reference issue #342 instead of #286.

**File:** `zb_io/src/extraction/patch/macos.rs` (+4/-20 lines)

Fixes #342

### DIFF
--- a/zb_io/src/extraction/patch/macos.rs
+++ b/zb_io/src/extraction/patch/macos.rs
@@ -128,27 +128,11 @@ fn patch_macho_binary_strings(path: &Path, new_prefix: &str) -> Result<(), Error
 
         if new_bytes.len() > old_bytes.len() {
             // Cannot expand shorter paths in-place in Mach-O binaries.
-            // Skip this prefix — the install_name_tool pass handles load
-            // command changes regardless of length, and many binaries
-            // legitimately reference shorter prefixes like /usr/local for
-            // system libraries (not Homebrew paths).
+            // Skip this old prefix — install_name_tool handles load command
+            // changes regardless of length, and text files are patched
+            // separately in the text-file pass.
             //
-            // See: https://github.com/lucasgelfond/zerobrew/issues/286
-            let has_old_paths = contents
-                .windows(old_bytes.len() + 1)
-                .any(|w| w[..old_bytes.len()] == *old_bytes && w[old_bytes.len()] == b'/');
-            if has_old_paths {
-                warn!(
-                    path = %path.display(),
-                    old_prefix = %old_prefix,
-                    new_prefix = %new_prefix,
-                    "binary contains hardcoded paths under {old_prefix} that \
-                    could not be rewritten to {new_prefix} (new path is longer). \
-                    this package may not work correctly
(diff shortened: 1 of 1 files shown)


cachebag (approved):
Thank you! Could you add a test or set of tests for this? I'd image we'd need something to ensure the root cause is resolved until we find a better solution.

PR #369, opened 2026-05-27
feat(zb_cli): `zb upgrade` to upgrade installed packages

pmnmq:

  # `zb upgrade`

  Adds the `zb upgrade` command and fixes a disk-leak bug in its underlying flow.

  ## What's new

  - **`zb upgrade [<formula>...]`** — upgrades installed packages. With no args, upgrades everything `zb outdated` would list; with args, upgrades only those.
  - **Flags**: `-s, --build-from-source` (forces source build, matching `zb install`), `--no-link` (skip symlinks).
  - Behavior on already-latest packages is a no-op (does not error).

  ## What's fixed

  The earlier in-tree `Installer::upgrade` implementation installed the new version but never removed the old version's `cellar/<pkg>/<old>/` directory or its `keg_files` rows. Repeated upgrades would
  accumulate orphan keg directories on disk until manual cleanup.

  The fix reorders the flow to **uninstall the old version first, then install the new one**, reusing the existing `uninstall_by_version` helper so all three cleanup steps (unlink symlinks, drop DB
  rows, decrement `store_refs`, remove cellar dir) happen as one unit. The whole operation runs under a single `install.lock` acquisition (a small split of `execute_with_progress` into a locked outer /
  inner pair makes this composable without re-entrant flock deadlock).

  ## Changes by commit

  | Commit | Purpose |
  |---|---|
  | `refactor(zb_io): drop unused link helpers and dedupe uninstall` | Removes dead `unlink_only` / `relink_only`; makes `uninstall` a thin wrapper over `uninstall_by_version` |
  | `fix(zb_io): clean up old cellar and \`keg_files\` rows on \`zb upgrade\`` | The actual bug fix: rewrites `Installer::upgrade` into its own module, holds `install.lock` across the full flow, calls
  `uninstall_by_version` before `execute_inner` |
  | `feat(zb_cli): \`zb upgrade\` to upgrade installed packages` | CLI plumbing — `Commands::Upgrade`, `commands/upgrade.rs`, dispatch |
  | `docs: document \`zb upgrade\` in README and CHANGELOG` | README quick-start entries (EN + zh) + CHANGELOG entry |

  ## Out of scope

  - **Cask upgrade** is not implemented; cask upgrade has no existing path in the codebase and adding one is a separate change. Passing a cask to `zb upgrade` falls through to the existing
  `plan_with_options` error path.
  - **Progress callback factory** between `install` and `upgrade` UI flows is not extracted; both keep their own inline `MultiProgress` setup for now. Reasonable to revisit once a third call site
  appears.

  ## Test plan

  - [x] `just build` clean (fmt-check → clippy → build)
  - [x] `just test` — 333 / 0

  ## Test plan

  - [x] `just build` clean (fmt-check → clippy → build)
  - [x] `just test` — 333 / 0
  - [x] Four new tests in `zb_io/src/installer/install/upgrade.rs`:
    - `upgrade_replaces_old_version_and_cleans_up` — asserts new cellar exists, **old cellar dir is gone**, symlink points at new version, DB version is updated
    - `upgrade_with_no_link_does_not_create_symlinks` — verifies `--no-link` semantics
    - `upgrade_no_op_when_already_latest` — idempotent when nothing to do
    - `upgrade_errors_when_not_installed` — surfaces `Error::NotInstalled`
  - [x] Existing `uninstall` tests (`uninstall_cleans_everything`, `uninstall_accepts_full_tap_reference_after_install`, `uninstalling_non_installed_tap_ref_does_not_remove_core_formula`) still pass —
  regression coverage for the dedup.

  #

  

  - [x] AI was used in this PR to generate code, documentation, or other changes, and I have explained _how_ it was used below. I understand that if I do not adhere to disclosing AI usage, my PR can be
  closed, even without explanation.

  ### AI usage disclosure

  Claude Code (Opus 4.7) was used as a pair-programming assistant for this PR. Specifically:

  - **Code review of an in-progress WIP** identified the cellar-leak bug and three pieces of unused helper code; surfaced the missing lock coverage and the structural mis-fit of the original
  implementation living in `mod.rs`.
  - **Implementation** of `zb_io/src/installer/install/upgrade.rs` (the new `Installer::upgrade` plus its 4 tests), the `execute_with_progress` / `execute_inner` split, and the `uninstall` dedupe
  refactor.
  - **Documentation** updates to `README.md` / `README.zh.md` / `CHANGELOG.md`.

  All design decisions (cleanup strategy: reuse `uninstall_by_version`; deferring cask upgrade and the shared progress-callback helper to follow-up work; commit-split granularity) were made by me after
  evaluating options the assistant proposed. Every change was manually reviewed and the test suite was run locally before each commit.

### DIFF
--- a/zb_cli/src/commands/upgrade.rs
+++ b/zb_cli/src/commands/upgrade.rs
@@ -1,0 +1,220 @@
+use console::style;
+use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+use std::time::Instant;
+use zb_io::{InstallProgress, ProgressCallback};
+
+use crate::ui::StdUi;
+use crate::utils::normalize_formula_name;
+
+pub async fn execute(
+    installer: &mut zb_io::Installer,
+    formulas: Vec<String>,
+    build_from_source: bool,
+    no_link: bool,
+    ui: &mut StdUi,
+) -> Result<(), zb_core::Error> {
+    let start = Instant::now();
+
+    let outdated = if formulas.is_empty() {
+        ui.heading("Checking for outdated packages...".to_string())
+            .map_err(ui_error)?;
+        let (outdated, warnings) = installer.check_outdated().await?;
+        for warning in &warnings {
+            eprintln!("{} {}", style("Warning:").yellow().bold(), warning);
+        }
+        if outdated.is_empty() {
+            ui.info("All packages are up to date.".to_string())
+                .map_err(ui_error)?;
+            return Ok(());
+        }
+        outdated
+    } else {
+        let mut normalized = Vec::with_capacity(formulas.len());
+        for formula in &formulas {
+            normalized.push(normalize_formula_name(formula)?);
+        }
+        let mut outdated = Vec::new();
+        for name in &normalized {
+            match installer.is_outdated(name).await {
+                Ok(Some(pkg)) => outdated.push(pkg),
+                Ok(None) => {
+                    ui.info(format!("{} is already up to date", name))
+                        .map_err(ui_error)?;
+                }
+                Err(zb_core::Error::NotInstalled { .. }) => {
+                    ui.error(format!("{} is not installed", name))
+                        .map_err(ui_error)?;
+                }
+                Err(e) => return Err(e),
+            }
+        }
+        if outdated.is_empty() {
+            ui.info("All specified packages are up to date.".to_string())
+                .map_err(ui_error)?;
+            return Ok(());
+        }
+        outdated
+    };
+
+    ui.heading(format!("Upgrading {}...", style(outdated.len()).bold()))
+        .map_err(ui_error)?;
+
+    let multi = MultiProgress::new();
+    let bars: Arc<Mutex<HashMap<String, ProgressBar>>> = Arc::new(Mutex::new(HashMap::new()));
+
+    let spinner_style = ProgressStyle::default_spinner()
+        .template("    {prefix:<16} {spinner:.cyan} {msg}")
+        .unwrap()
+        .tick_chars("⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏");
+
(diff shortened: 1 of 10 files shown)


zb_io/src/installer/install/upgrade.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        // Empty plan = already on the latest version. No-op, keep installation.
+        if plan.items.is_empty() {
+            return Ok(());
+        }
+
+        // Clear old version first: unlinks symlinks, removes the keg directory,
+        // drops `installed_kegs` / `keg_files` rows, decrements `store_refs`.
+        // After this the new install path is unobstructed.
+        self.uninstall_by_version(name, &old.version)?;
```
cachebag, line 53: Can we stage/download/build the new version first, then swap links/remove the old keg only after success, or restore the old version on failure? This also needs a failure-path test.

zb_cli/src/commands/upgrade.rs
```diff
@@ -0,0 +1,220 @@
...
+        }
+        let mut outdated = Vec::new();
+        for name in &normalized {
+            match installer.is_outdated(name).await {
+                Ok(Some(pkg)) => outdated.push(pkg),
+                Ok(None) => {
+                    ui.info(format!("{} is already up to date", name))
+                        .map_err(ui_error)?;
+                }
+                Err(zb_core::Error::NotInstalled { .. }) => {
```
cachebag, line 50: For an explicitly requested package this should make the command fail, not just print and continue. As written, `zb upgrade does-not-exist` exits 0 and then reports “All specified packages are up to date”; with mixed args it also hides the missing package if the rest succeed.

cachebag (changes requested):
Please also address copilot's comments as they are valid.

Thank you!

pmnmq:
> Please also address copilot's comments as they are valid.
> 
> Thank you!

Thanks for your review! I will refine the PR based on your suggestions.

cachebag (approved):

PR #383, opened 2026-05-30
fix(zb_io): skip libexec/ for Python virtualenv kegs

### LINKED ISSUE #377: link conflict when installing multiple Python CLI apps sharing virtualenv entry points
### Description

  Installing two Python CLI packages sequentially (e.g. `mycli` then `pgcli`) fails at the link step with a conflict error, even though
  both packages downloaded and installed successfully.

  ### Steps to Reproduce

```shell
zb install mycli
zb install pgcli
```

```shell
~/p/o/zerobrew ❯❯❯ zb install mycli
==> Installing mycli...
==> Resolving dependencies (16 packages)...
    • ca-certificates 2026-05-14
    • certifi 2026.5.20
    • libyaml 0.2.5
    • lz4 1.10.0
    • mpdecimal 4.0.1
    • openssl@3 3.6.2
    • pycparser 3.0
    • cffi 2.0.0
    • cryptography 48.0.0
    • pydantic 2.13.4
    • readline 8.3.3
    • sqlite 3.53.1
    • xz 5.8.3
    • zstd 1.5.7
    • python@3.14 3.14.5
    • mycli 1.73.0
==> Downloading and installing formulas...

==> Installed 16 packages in 1.81s
~/p/o/zerobrew ❯❯❯ zb install pgcli
==> Installing pgcli...
==> Resolving dependencies (13 packages)...
    • ca-certificates 2026-05-14
    • icu4c@78 78.3
    • lz4 1.10.0
    • mpdecimal 4.0.1
    • openssl@3 3.6.2
    • krb5 1.22.2
    • libpq 18.4
    • readline 8.3.3
    • sqlite 3.53.1
    • xz 5.8.3
    • zstd 1.5.7
    • python@3.14 3.14.5
    • pgcli 4.4.0
==> Downloading and installing formulas...

error: The link step did not complete successfully.
The formula was installed, but is not symlinked into the prefix.

Possible conflicting files:
  /opt/zerobrew/libexec/bin/sqlformat (symlink belonging to mycli)
  /opt/zerobrew/libexec/bin/pygmentize (symlink belonging to mycli)
  /opt/zerobrew/libexec/bin/tabulate (symlink belonging to mycli)

error: link conflicts:
  '/opt/zerobrew/libexec/bin/sqlformat' (owned by mycli)
  '/opt/zerobrew/libexec/bin/pygmentize' (owned by mycli)
  '/opt/zerobrew/libexec/bin/tabulate' (owned by mycli)
~/p/o/zerobrew ❯❯❯
```
  ### Root Cause

  Homebrew bottles for Python apps place all dependencies inside an isolated libexec/ virtualenv. The main executable is exposed via a bin/
   symlink (e.g. bin/mycli → ../libexec/bin/mycli), while shared dependency entry points like sqlformat, pygmentize, and tabulate live only
   inside libexec/bin/ and are not intended to be merged into the shared prefix.

  zerobrew's linker merges all of LINK_DIRS (including libexec/) into the shared prefix for every keg. This causes
  prefix/libexec/bin/sqlformat to be owned by the first package installed, and when any subsequent package sharing the same Python
  dependencies is installed, the linker raises a conflict.

  Homebrew avoids this by not linking libexec/ into the shared prefix for Python virtualenv kegs — access to the package is provided
  entirely through the bin/ symlinks.

  ### Workaround

  None currently. The second package remains installed but unlinked and is therefore not usable.

  ### Environment

  - OS: macOS (Apple Silicon)
  - zerobrew version: 0.2.1
  - Affected combinations: any two Python CLI apps sharing PyPI dependencies (e.g. mycli + pgcli, ranger + ansible-lint)

pmnmq:
#

Claude Code 4.7 was used as a pair-programming assistant for this PR. Specifically:

- **Problem localization**: helped identify that the conflict originated in `libexec/bin/` subdirectories rather than top-level
`libexec/` metadata, and pointed to the relevant section of `link.rs`.
- **Code review**: reviewed the implementation and test coverage after the fix was written.

## What's fixed

Fixes #377

Installing two Python CLI apps in sequence (e.g. `zb install mycli && zb install pgcli`) fails with a link conflict:

  error: link conflicts:
    '/opt/zerobrew/libexec/bin/sqlformat' (owned by mycli)
    '/opt/zerobrew/libexec/bin/pygmentize' (owned by mycli)
    '/opt/zerobrew/libexec/bin/tabulate'   (owned by mycli)

  Homebrew Python CLI formulas bundle an isolated virtualenv under `libexec/`. The main executable is exposed only via a `bin/` symlink
  that resolves into `libexec/` within the keg itself — `libexec/bin/` is never intended to be merged into the shared prefix. Shared
  transitive dependencies (sqlformat, pygmentize, etc.) appear in `libexec/bin/` for every Python app, so as soon as two are installed they
   collide.

  The prior partial fix skipped `pyvenv.cfg` and `.gitignore` directly inside `libexec/` but did not skip subdirectories, so the actual
  conflict path (`libexec/bin/`) was still being linked.

  The fix detects Python virtualenv kegs by the presence of `libexec/pyvenv.cfg` and skips the entire `libexec/` directory in both
  `check_conflicts` and `link_keg` — matching Homebrew's own behavior. `unlink_keg` and `collect_linked_files` are intentionally left
  unchanged so that existing installs made before this fix can still be cleaned up correctly.

  Also extracted `LIBEXEC_DIR` and `PYVENV_CFG` as named constants to avoid repeating the string literals across the detection logic and
  the skip list.

  ## Out of scope

  - **Kegs already installed before this fix** retain their `prefix/libexec/bin/` symlinks until `zb uninstall` is run. This is intentional
   — `unlink_keg` still iterates `libexec/` for backward compatibility.
  - **Non-Python `libexec/` usage** (e.g. git's `libexec/git-core/`) is unaffected — the guard only fires when `pyvenv.cfg` is present.

  ## Test plan

  - [x] `just build` clean (fmt-check → clippy → build)
  - [x] `just test` — 330 / 0
  - [x] Strengthened existing `skips_libexec_virtualenv_metadata_to_avoid_conflicts`: now also covers `libexec/bin/` subdirectories, not
  just top-level metadata files
  - [x] New `python_virtualenv_keg_does_not_link_libexec_bin` — single Python venv keg: `prefix/bin/mycli` exists,
  `prefix/libexec/bin/sqlformat` does not
  - [x] New `two_python_virtualenv_kegs_with_shared_dep_names_install_without_conflict` — the exact mycli + pgcli regression: both
  `link_keg` calls succeed, shared dep entry points are absent from the prefix
  - [x] Existing `links_libexec_directory` (non-Python keg with `libexec/git-core/`) still passes

## Verification
Before:
截屏2026-05-30 19 31 00

After:
截屏2026-05-30 19 32 54

### DIFF
--- a/zb_io/src/cellar/link.rs
+++ b/zb_io/src/cellar/link.rs
@@ -733,10 +733,11 @@ mod tests {
 
     #[test]
     fn two_python_virtualenv_kegs_with_shared_dep_names_install_without_conflict() {
-        // Regression for #377: installing two Python CLI apps (mycli, pgcli)
-        // failed because shared transitive deps (sqlformat, pygmentize) live in
-        // each keg's libexec/bin/ and previously got merged into the shared
-        // prefix/libexec/bin/, colliding on the second install.
+        // Regression test for #377 (https://github.com/lucasgelfond/zerobrew/issues/377):
+        // installing two Python CLI apps (mycli, pgcli) failed because shared
+        // transitive deps (sqlformat, pygmentize) live in each keg's libexec/bin/
+        // and previously got merged into the shared prefix/libexec/bin/, colliding
+        // on the second install.
         let tmp = TempDir::new().unwrap();
         let prefix = tmp.path();
         let linker = Linker::new(prefix).unwrap();
--- a/zb_io/src/cellar/link.rs
+++ b/zb_io/src/cellar/link.rs
@@ -5,18 +5,36 @@ use std::path::{Component, Path, PathBuf};
 use zb_core::{ConflictedLink, Error};
 
 const LINK_DIRS: &[&str] = &["bin", "lib", "libexec", "include", "share", "etc"];
-const LIBEXEC_SKIP_FILES: &[&str] = &[".gitignore", "pyvenv.cfg"];
+const PYVENV_CFG: &str = "pyvenv.cfg";
+const LIBEXEC_SKIP_FILES: &[&str] = &[".gitignore", PYVENV_CFG];
 
 fn should_skip_link_entry(src_dir: &Path, entry_name: &std::ffi::OsStr) -> bool {
-    // Homebrew-style Python virtualenv formulae commonly place metadata files at
-    // libexec/.gitignore, libexec/pyvenv.cfg, and private site-packages trees.
-    // Linking these into a shared prefix/libexec causes cross-formula conflicts
-    // even though they are private virtualenv contents, not user entrypoints.
-    (src_dir.file_name().and_then(|n| n.to_str()) == Some("libexec")
-        && entry_name
+    // Homebrew-style Python virtualenv formulae bundle an isolated venv under
+    // libexec/. The main executable is exposed via bin/<name> symlinks that
+    // resolve into libexec/ within the keg itself, so nothing inside libexec/
+    // is meant for the shared prefix. Linking libexec/ contents into the
+    // shared prefix/libexec/ causes cross-formula conflicts:
+    //   - libexec/{pyvenv.cfg, .gitignore} (metadata)
+    //   - libexec/lib*/python3.X/site-packages/ (private dep trees)
+    //   - libexec/bin/<shared-dep> (e.g. sqlformat across mycli + pgcli)
+    let is_libexec_dir = src_dir.file_name().and_then(|n| n.to_str()) == Some("libexec");
+
+    if is_libexec_dir {
+        // Detect a virtualenv libexec by the presence of pyvenv.cfg alongside;
+        // when present, skip every entry — including libexec/bin/ — so private
+        // venv contents never leak into the shared prefix.
+        if src_dir.join(PYVENV_CFG).exists() {
+            return true;
+        }
+        if entry_name
             .to_str()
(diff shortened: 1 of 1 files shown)


zb_io/src/cellar/link.rs
```diff
@@ -688,6 +731,53 @@ mod tests {
         );
     }
 
+    #[test]
+    fn two_python_virtualenv_kegs_with_shared_dep_names_install_without_conflict() {
+        // Regression for #377: installing two Python CLI apps (mycli, pgcli)
```
cachebag:
```suggestion
        // Regression test for #377: installing two Python CLI apps (mycli, pgcli)
```
Feel a bit weird about this wording, so I suggest saying this. 

This implies #377's original fix regressed which isn't really true, it just did not cover this behavior. Could you also add a direct link to the github issue here?

cachebag (approved):
LGTM modulo one nit

zb_io/src/cellar/link.rs
pmnmq: Thanks for the suggestion, I'll update it.

